PR #1, opened 2017-04-05
fix up the pre-commit hook

richvdh:
* Run the checks against the git index (ie, what you're about to commit), rather than the working copy

* Add the changes made by `go fmt` back to the index, so that they are included in the commit, as well as the working copy.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -3,8 +3,8 @@
 set -eu
 
 # make the GIT_DIR and GIT_INDEX_FILE absolute, before we change dir
-export GIT_DIR=$(realpath `git rev-parse --git-dir`)
-export GIT_INDEX_FILE=$(realpath `git rev-parse --git-path index`)
+export GIT_DIR=$(readlink -f `git rev-parse --git-dir`)
+export GIT_INDEX_FILE=$(readlink -f `git rev-parse --git-path index`)
 
 wd=`pwd`
 
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -4,7 +4,9 @@ set -eu
 
 # make the GIT_DIR and GIT_INDEX_FILE absolute, before we change dir
 export GIT_DIR=$(readlink -f `git rev-parse --git-dir`)
-export GIT_INDEX_FILE=$(readlink -f `git rev-parse --git-path index`)
+if [ -n "${GIT_INDEX_FILE:+x}" ]; then
+    export GIT_INDEX_FILE=$(readlink -f "$GIT_INDEX_FILE")
+fi
 
 wd=`pwd`
 
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,9 +1,41 @@
-#! /bin/bash
+#!/bin/bash
 
 set -eu
 
+# make the GIT_DIR and GIT_INDEX_FILE absolute, before we change dir
+export GIT_DIR=$(realpath `git rev-parse --git-dir`)
+export GIT_INDEX_FILE=$(realpath `git rev-parse --git-path index`)
+
+wd=`pwd`
+
+# create a temp dir
+tmpdir=`mktemp -d`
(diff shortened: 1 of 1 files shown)


hooks/pre-commit
```diff
@@ -1,9 +1,41 @@
...
 
+# make the GIT_DIR and GIT_INDEX_FILE absolute, before we change dir
+export GIT_DIR=$(realpath `git rev-parse --git-dir`)
+export GIT_INDEX_FILE=$(realpath `git rev-parse --git-path index`)
+
+wd=`pwd`
+
+# create a temp dir
+tmpdir=`mktemp -d`
+trap 'rm -rf "$tmpdir"' EXIT
```
Kegsay: What does this do? I'm guessing it re-creates it if it already existed or something? Can you add this to the comment on `:11` please?

hooks/pre-commit
```diff
@@ -1,9 +1,41 @@
...
+
+# we now need to apply any changes made to both the index and the working copy.
+# so, first get a patch
+git diff > "$GIT_DIR/pre-commit.patch"
+
+# add the changes to the index
+git add -u
+
+# attempt to apply the changes to the wc (but don't fail the commit for it if
+# there are conflicts).
```
Kegsay: Why shouldn't we fail the commit if there are conflicts? Won't it be bad?

hooks/pre-commit
richvdh: done 22ce003

hooks/pre-commit
richvdh: well, the commit itself will be ok; the only thing is that the changes made by `go fmt` won't be replicated to your working copy.

Suppose you have `git add`ed a file, and then made some more changes but not added them. When the pre-commit hook runs, we run `go fmt` against the staged version before it is committed; the problem is that the changes `go fmt` makes could conflict with the later changes you made in your working copy. I'm not sure what the correct behaviour would be in those circumstances, but I'd rather leave the working copy alone than mess it up. On the other hand, I see no reason that this should block the commit.

hooks/pre-commit
```diff
@@ -1,9 +1,42 @@
-#! /bin/bash
+#!/bin/bash
```
Kegsay: Can you please briefly:
 - Describe the problem with the previous "obvious" pre-commit hook.
 - Outline how your new code fixes this problem.

We're doing a lot of extra stuff now, and I'm happy for it provided we justify why we should be doing so. It will also make it possible for us to fix this problem in a different way in the future potentially. At present, you merely state what you're doing, not why.

hooks/pre-commit
Kegsay: By blocking the commit, it sounds like you're punting the conflict resolution until later. I don't know if that is a good or bad thing. I'm tempted to suggest erring on the side of failing fast (that is, forcing the user to resolve the conflict there and then), mainly so the user can see cause and effect more quickly. WDYT?

Kegsay:
Travis still failing.

hooks/pre-commit
richvdh: I don't think this will be a problem in practice: the conflict will sort itself out on the next run of `go fmt` in the working copy (or once those later changes are committed). 

I do think that blocking the commit would be frustrating, though.

Can we try it like this and see what happens?

hooks/pre-commit
richvdh: It would certainly have helped if I hadn't thinkoed when writing the commit comment/PR description. I have updated the PR description and improved a comment in 77ca493; is it better now?

richvdh:
> Travis still failing.

indeed; trying to figure out what is wrong with it

richvdh:
assuming travis now sorts its life out, @Kegsay: PTAL

hooks/pre-commit
Kegsay: Sure, WFM.

Kegsay:
Comments look fab, so provided travis is happy, LGTM.

PR #3, opened 2017-04-05
Serve log files as plain text

richvdh:
If the client accepts gzip, just serve the file as a gzip-encoded plain text.

If it doesn't, gunzip it first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,14 +5,13 @@ Web service which collects and serves bug reports.
 To run it, do:
 
 ```
-BUGS_USER=<user> BUGS_PASS=<password> go run src/github.com/matrix-org/rageshake/main.go PORT
+go get github.com/constabulary/gb/...
+gb build
+BUGS_USER=<user> BUGS_PASS=<password> ./bin/rageshake PORT
+# example:
+# BUGS_USER=alice BUGS_PASS=secret ./bin/rageshake 8080
 ```
 
-Example:
-
-```
-BUGS_USER=alice BUGS_PASS=secret go run src/github.com/matrix-org/rageshake/main.go 8080
-```
 
 ## HTTP endpoints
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -5,14 +5,13 @@ Web service which collects and serves bug reports.
 To run it, do:
 
 ```
-BUGS_USER=<user> BUGS_PASS=<password> go run src/github.com/matrix-org/rageshake/main.go PORT
+go get github.com/constabulary/gb/...
+gb build
```
Kegsay: FWIW I see no reason why this should be a GB project - we don't actually depend on anything that isn't in the stdlib. I don't expect you to change this now, but it may be nice to change this in the future.

src/github.com/matrix-org/rageshake/logserver.go
```diff
@@ -0,0 +1,163 @@
...
+		r.URL.Path = upath
+	}
+
+	log.Println("Serving", upath)
+
+	// eliminate ., .., //, etc
+	upath = path.Clean(upath)
+
+	// reject some dodgy paths
+	if containsDotDot(upath) || strings.Contains(upath, "\x00") || (filepath.Separator != '/' && strings.IndexRune(upath, filepath.Separator) >= 0) {
```
Kegsay: This makes me uneasy because getting this wrong could introduce a security vulnerability. You need to convince me that these checks are sufficient. If you nabbed this from the `FileServer` directly, then great, please add a line comment referencing the line number so I can check that these checks are equivalent. If you didn't nab this from the `FileServer`, please justify it.

Kegsay:
Otherwise this looks very very sensible. LGTM.

README.md
richvdh: well, I'm about to add a dep on https://godoc.org/github.com/google/go-github, which changes that.

src/github.com/matrix-org/rageshake/logserver.go
richvdh: have added a comment

README.md
Kegsay: Oh okay.

Kegsay:
LGTM

PR #5, opened 2017-04-10
Switch to yaml for our config data

richvdh:
Env vars are getting unwieldy, and it's going to get worse.

### DIFF
--- a/src/github.com/matrix-org/rageshake/main.go
+++ b/src/github.com/matrix-org/rageshake/main.go
@@ -19,13 +19,29 @@ package main
 import (
 	"context"
 	"crypto/subtle"
+	"flag"
 	"fmt"
 	"github.com/google/go-github/github"
 	"golang.org/x/oauth2"
+	"io/ioutil"
 	"log"
 	"net/http"
 	"os"
+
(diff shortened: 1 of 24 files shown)


src/github.com/matrix-org/rageshake/main.go
```diff
@@ -69,17 +90,29 @@ func main() {
...
+func loadConfig(configPath string) (*config, error) {
+	contents, err := ioutil.ReadFile(configPath)
+	if err != nil {
+		return nil, err
+	}
+	var cfg config
+	if err = yaml.Unmarshal(contents, &cfg); err != nil {
+		return nil, err
+	}
+	// check required fields
```
Kegsay: There are no required fields.

Kegsay:
Other than comment on a comment, LGTM!

src/github.com/matrix-org/rageshake/main.go
richvdh: Thanks. Maybe you should also fix the place I cargo-culted this from: https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/cmd/dendrite-sync-server/main.go#L47 :)

PR #6, opened 2017-04-10
Add 'app' and 'data' fields to report API

richvdh:


src/github.com/matrix-org/rageshake/submit.go
```diff
@@ -41,10 +41,12 @@ type submitServer struct {
...
-	Text      string     `json:"text"`
-	Version   string     `json:"version"`
-	UserAgent string     `json:"user_agent"`
-	Logs      []logEntry `json:"logs"`
+	Text      string            `json:"text"`
+	AppName   string            `json:"app"`
+	Version   string            `json:"version"`
+	UserAgent string            `json:"user_agent"`
+	Logs      []logEntry        `json:"logs"`
+	Data      map[string]string `json:"data"`
```
Kegsay: > `data`: a set of arbitrary name/value pairs

If so, ITYM `map[string]interface{}` otherwise the server will fail to decode numbers, arrays, objects, etc.

src/github.com/matrix-org/rageshake/submit.go
```diff
@@ -93,12 +95,20 @@ func (s *submitServer) saveReport(ctx context.Context, p payload) error {
...
+
+	var summaryBuf bytes.Buffer
+	fmt.Fprintf(
+		&summaryBuf,
+		"%s\n\nNumber of logs: %d\nApplication: %s\nVersion: %s\nUser-Agent: %s\n",
+		p.Text, len(p.Logs), p.AppName, p.Version, p.UserAgent,
 	)
-	if err := gzipAndSave([]byte(summary), prefix, "details.log.gz"); err != nil {
+	for k, v := range p.Data {
+		fmt.Fprintf(&summaryBuf, "%s: %s\n", k, v)
```
Kegsay: Assuming https://github.com/matrix-org/rageshake/pull/6/files#r110698886 is true, you'll need to change this too.

src/github.com/matrix-org/rageshake/submit.go
```diff
@@ -93,12 +95,20 @@ func (s *submitServer) saveReport(ctx context.Context, p payload) error {
 	//  "bugreport-20170115-112233-N.log.gz" => oldest log
 	t := time.Now().UTC()
 	prefix := t.Format("2006-01-02/150405")
-	summary := fmt.Sprintf(
-		"%s\n\nNumber of logs: %d\nVersion: %s\nUser-Agent: %s\n", p.Text, len(p.Logs), p.Version, p.UserAgent,
+
+	var summaryBuf bytes.Buffer
+	fmt.Fprintf(
+		&summaryBuf,
+		"%s\n\nNumber of logs: %d\nApplication: %s\nVersion: %s\nUser-Agent: %s\n",
```
Kegsay: I'm surprised this actually works as intended. Yes, you should use `bytes.Buffer` to efficiently concatenate strings, but there is a specific `buf.WriteString(str)` function you should use, and then `buf.String()` to get the complete output. That would be much clearer, though you'll still need to `fmt.Sprintf` to format the string.

src/github.com/matrix-org/rageshake/submit.go
richvdh: hrm. I'm not keen on letting it be arrays/objects, since all we end up doing is figuring out how to turn it into a string. It would be nice if it stringified numbers automatically: any idea if there's a neat way to achieve this?

src/github.com/matrix-org/rageshake/submit.go
richvdh: bytes.Buffer is a writer, so why can't I write to it? Why bother going through a string for the formatting?

As for using `buf.String()`: I want to end up with the bytes (to stick into `gzipAndSave`), so again, why bother turning it into an intermediate string?

src/github.com/matrix-org/rageshake/submit.go
Kegsay: > neat way to achieve this

This is Go. Of course there isn't a neat way. You may be able to abuse `json.RawMessage` and then just read the literal `[]byte` given it's JSON. E.g: https://play.golang.org/p/-b3BpeSBgf

src/github.com/matrix-org/rageshake/submit.go
Kegsay: IMO it's cognitively easier to grok what:
```go
summaryBuf.WriteString(fmt.Sprintf("Hello %s", name))
```
is doing than:
```go
fmt.Fprintf(&summaryBuf, "Hello %s", name)
```

As for `buf.String()` - if you want a byte output then fair enough, for some reason I thought you wanted a string.

src/github.com/matrix-org/rageshake/submit.go
richvdh: ftsd. I'm sticking with strings. I could update the readme if you think it's merited.

src/github.com/matrix-org/rageshake/submit.go
richvdh: IMO it's not :/

src/github.com/matrix-org/rageshake/submit.go
Kegsay: Please update the readme then: clarity trumps all, and at the moment it's a bit ambiguous:

> arbitrary name/value pairs

src/github.com/matrix-org/rageshake/submit.go
Kegsay: Okay then, I'm happy to ignore this then.

Kegsay:
Aside from README update, LGTM!

PR #7, opened 2017-04-10
Make github issue a bit more useful

richvdh:
Put stuff in the title and body of the github issue, and pick a repo based on the application specified in the API call.

### DIFF
--- a/src/github.com/matrix-org/rageshake/main.go
+++ b/src/github.com/matrix-org/rageshake/main.go
@@ -29,6 +29,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	yaml "gopkg.in/yaml.v2"
 )
@@ -84,6 +85,7 @@ func main() {
 			&oauth2.Token{AccessToken: cfg.GithubToken},
 		)
 		tc := oauth2.NewClient(ctx, ts)
+		tc.Timeout = time.Duration(5) * time.Minute
 		ghClient = github.NewClient(tc)
 	}
 
--- a/src/github.com/matrix-org/rageshake/main.go
+++ b/src/github.com/matrix-org/rageshake/main.go
@@ -46,6 +46,8 @@ type config struct {
 
 	// A GitHub personal access token, to create a GitHub issue for each report.
 	GithubToken string `yaml:"github_token"`
+
+	GithubProjectMappings map[string]string `yaml:"github_project_mappings"`
 }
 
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/rageshake/submit.go
```diff
@@ -38,6 +39,11 @@ type submitServer struct {
 	// github client for reporting bugs. may be nil, in which case,
 	// reporting is disabled.
 	ghClient *github.Client
+
+	APIPrefix string
+
+	// mappings from application to github owner/project
+	GithubProjectMappings map[string]string
```
Kegsay: There's no real distinction here between public and private given it's all in the same package and isn't used outside of it (fun fact: it *can't* be used outside because you cannot export `package main`). I'd keep it all private.

Also, one-liner on API prefix would be schweet please.

src/github.com/matrix-org/rageshake/submit.go
```diff
@@ -117,15 +128,46 @@ func (s *submitServer) saveReport(ctx context.Context, p payload) error {
...
+		log.Println("Can't create GH issue for invalid repo", ghProj)
+	}
+	owner, repo := splits[0], splits[1]
+
+	var title string
+	if userText == "" {
+		title = "Untitled report"
+	} else {
+		// set the title to the first line of the user's report
+		if i := strings.IndexAny(userText, "\r\n"); i < 0 {
```
Kegsay: And what if the user is using `\n` delimiters only? I would probably cheat and just do https://play.golang.org/p/UgsoFkP9tP

Kegsay: Ohhhh, `IndexAny` takes a list of *characters*. That's disappointing. Never mind, carry on then.

src/github.com/matrix-org/rageshake/main.go
```diff
@@ -80,7 +87,20 @@ func main() {
 		ghClient = github.NewClient(tc)
```
Kegsay: `tc.Timeout = 5 minutes` or something please.

Kegsay:
Otherwise LGTM. Merge after changes please.

src/github.com/matrix-org/rageshake/submit.go
richvdh: does this apply to the `config` structure too?

richvdh: assuming not, for now

PR #8, opened 2017-04-12
Rename sample config file

richvdh:
This means that we can have a `rageshake.yaml` on the live server which is used
without faffing with parameters.


Kegsay:
LGTM

PR #9, opened 2017-04-13
Undo the android client's mangled submissions

richvdh:
current riot-android stuffs everything into the 'version' field; I don't really
want that ending up in the github issue (especially the mxid), and I don't want
to have to wait for an update to riot-android to land, so let's unpick
it into 'data'.

### DIFF
--- a/src/github.com/matrix-org/rageshake/submit_test.go
+++ b/src/github.com/matrix-org/rageshake/submit_test.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2017 Vector Creations Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


src/github.com/matrix-org/rageshake/submit_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+// check that we can unpick the json submitted by the android clients
+func TestUnpickAndroidMangling(t *testing.T) {
+	body := `{"text": "test ylc 001",
+"version": "User : @ylc8001:matrix.org\nPhone : Lenovo P2a42\nVector version: 0:6:9\n",
+"user_agent": "Android"
+}`
+	p := testParsePayload(t, body, "")
+	if p.Text != "test ylc 001" {
+		t.Error("unexpected user text", p.Text)
```
Kegsay: The general Go idiom for test failure messages of the form:
```
terse message: got %s, want %s
```
or
```
funcName(args, here): got %s, want %s
```

We should be doing the same.

Kegsay:
Otherwise LGTM

PR #10, opened 2017-04-13
Send github URL back to application

richvdh:
I'm not sure if this is going to be useful, but it's fairly harmless, and is
annoying me by being on a branch.


Kegsay:
LGTM

PR #11, opened 2017-04-18
Zero-pad logfile names

richvdh:
- so that they are sorted properly


Kegsay:
LGTM - didn't even know that was possible in `fmt`!

PR #14, opened 2017-04-27
Add support for attaching files to reports

richvdh:
If a 'file' parameter is provided, save it as a file on disk. Also include a link to it from the github issue.

### DIFF
--- a/src/github.com/matrix-org/rageshake/submit.go
+++ b/src/github.com/matrix-org/rageshake/submit.go
@@ -83,13 +83,32 @@ func (s *submitServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	p := parseRequest(w, req)
-	if p == nil {
-		// parseRequest already wrote an error
+	// create the report dir before parsing the request, so that we can dump
+	// files straight in
+	t := time.Now().UTC()
+	prefix := t.Format("2006-01-02/150405")
+	reportDir := filepath.Join("bugs", prefix)
+	if err := os.MkdirAll(reportDir, os.ModePerm); err != nil {
+		log.Println("Unable to create report directory", err)
+		http.Error(w, "Internal error", 500)
 		return
 	}
 
-	resp, err := s.saveReport(req.Context(), *p)
+	listingURL := s.apiPrefix + "/listing/" + prefix
+	log.Println("Handling report submission; listing URI will be", listingURL)
+
+	p := parseRequest(w, req)
+	if p == nil {
+		// parseRequest already wrote an error, but now let's delete the
+		// useless report dir
+		if err := os.RemoveAll(reportDir); err != nil {
+			log.Printf("Unable to remove report dir %s after invalid upload: %v\n",
+				reportDir, err)
+		}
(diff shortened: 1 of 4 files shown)


richvdh:
@NegativeMjark could you take a look at this in kegan's absence? The two commits should stand alone; hopefully it is fairly obvious what is going on.

NegativeMjark:
Hmm, where are the files served from? What protection do we have against XSS here?
Are all the files served with a `text/plain` MIME-type or something?

src/github.com/matrix-org/rageshake/submit.go
```diff
@@ -245,19 +277,39 @@ func parseFormPart(part *multipart.Part, p *payload) error {
...
+func saveFormPart(originalFileName string, reader io.Reader, reportDir string) (string, error) {
+	// figure out a safe name. Replace slashes, other path separators, and
+	// leading dots with underscores.
+	leafName := strings.Replace(originalFileName, "/", "_", -1)
+	leafName = strings.Replace(leafName, string(filepath.Separator), "_", -1)
+	if leafName == "" {
+		leafName = "unnamed-file"
+	} else if leafName[0] == '.' {
+		leafName = "_" + leafName[1:]
+	}
```
NegativeMjark: Hmm, I'd be a lot happier if we URI encoded the originalFileName here, or used a random ID to store the file on disk. Allowing users to supply file names that contain either special shell characters or terminal escape characters or invalid utf8 could be saddening.

richvdh:
> Hmm, where are the files served from? What protection do we have against XSS here?
Are all the files served with a text/plain MIME-type or something?

interesting point. They are served from https://riot.im/bugreports/listing/, via `http.ServeFile`, which picks a mime-type based on the file extension. So this is probably an XSS attack waiting to happen.

The problem is that I want to be able to use this for attaching screenshots and things, so a fixed mime-type isn't going to cut it. I suppose I could limit it to a set of known safe extensions and treat everything else as octet-stream.

NegativeMjark:
So it looks like we serve the files here using `net/http.ServeFile`? https://github.com/matrix-org/rageshake/blob/master/src/github.com/matrix-org/rageshake/logserver.go#L82

It looks like `net/http.ServeFile` eventually calls the same code as `net/http.ServeContent`.

I think `net/http.ServeContent` will try to set the Content-Type using `mime.TypesByExtension`, if that fails then it will try to set it by guessing from the file contents using `net/http.DetectContentType`.

I assume you are guarding against XSS by restricting the list of allowed file extensions and the relying on `mime.TypesByExtension` to give a safe file type.

According to https://golang.org/pkg/mime/#TypeByExtension it will use a builtin list of file types falling back to the mime type lists in /etc.

From looking at the code the builtin list includes `.jpg` and `.png` but has no entry for `.txt`. I'd feel a lot better if you explicitly registered `.txt`, `.jpg` and `.png` using `mime.AddExtensionType` and added some comments to explain how we were protecting against XSS here.

Also might be worth setting a `Content-Security-Policy` or something?

richvdh:
@NegativeMjark ptal

PR #16, opened 2017-05-02
	Support for adding labels to github issues

richvdh:
... to make triaging easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,11 @@ logs.)
 
 * `version`: Application version. Included in the `details.log.gz` file.
 
+* `label`: Label to attach to the github issue, and include in the details file.
+
+  If using the JSON upload encoding, this should be encoded as a `labels` field,
+  whose value should be a list of strings.
+
 * `log`: a log file, with lines separated by newline characters. Multiple log
   files can be included by including several `log` parts.
 
--- a/src/github.com/matrix-org/rageshake/submit_test.go
+++ b/src/github.com/matrix-org/rageshake/submit_test.go
@@ -228,10 +228,6 @@ func checkParsedMultipartUpload(t *testing.T, p *payload) {
 }
 
 func stringSlicesEqual(got, want []string) bool {
-	if (got == nil) != (want == nil) {
-		return false
-	}
-
 	if len(got) != len(want) {
 		return false
 	}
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -49,6 +49,11 @@ logs.)
 
 * `version`: Application version. Included in the `details.log.gz` file.
 
+* `label`: Label to attach to the github issue, and include in the details file.
+
+  If using the JSON upload encoding, this should be encoded as a `labels` field,
```
Kegsay: It's bad enough that we have both JSON and multipart support, I'd rather we didn't diverge any more than we have to. Why does the JSON form support multiple labels on the GH issue, but the multipart form is strictly limited to 1 label? Seems bizarre.

Kegsay: Oh I see, each form part represents 1 label. I guess it's easier to do it like that than add in string splitting code.

Kegsay:
LGTM - I assume the check failures is because it depends on #15

richvdh:
the check failures are I think because the tests fail the cyclomatic complexity test, which is fixed in #14.

richvdh:
Sigh; merging #14 had the opposite effect.

@Kegsay: PTAL at the latest tweaks. In particular I wonder if using something like `reflect.DeepEqual` would be better than defining my own array comparator.

src/github.com/matrix-org/rageshake/submit_test.go
```diff
@@ -215,6 +227,23 @@ func checkParsedMultipartUpload(t *testing.T, p *payload) {
 	}
 }
 
+func stringSlicesEqual(got, want []string) bool {
+	if (got == nil) != (want == nil) {
```
Kegsay: `nil` slices have a length of 0, so unless you really really care about the distinction between a `nil` slice and an empty slice, I would skip this check and rely on `len()` instead. `nil` slices are also `range`able so in the case where they are both 0 then the `for` loop body won't execute. If we do care about the distinction e.g there is a subtle semantic difference in your code between the two, I'd rather we modified the code such that it didn't matter.

Kegsay:
>  In particular I wonder if using something like `reflect.DeepEqual` would be better than defining my own array comparator

In this particular case, I don't care which you use. Your intent is obvious either way IMO.

ara4n:
does this mean we can merge the 3 rageshake repos into one, and use labels to demux the issues? O:-)

Kegsay:
LGTM

richvdh:
> does this mean we can merge the 3 rageshake repos into one, and use labels to demux the issues? O:-)

if you want, but the reason for keeping them separate was for better email notification support.

PR #17, opened 2017-05-03
Save uploaded logs during upload

richvdh:
I got fed up with shoving everything into the structure the JSON api uses for
upload, so now parseRequest returns its own structure, and saves the log files
to disk in the same way as the other attachments.

This also means that we can test the saving of log files more representatively.


Kegsay:
I'm not really seeing the benefit to doing this (size delta for the non-test file is `+101 -55`, clarity doesn't seem to be improved, no real performance change from what I can see, doesn't add any new support for anything), so I'd be interested to know what made you fed up.

Code-wise, LGTM.

PR #18, opened 2017-05-03
If a log is uploaded with a sensible filename, preserve it

richvdh:
the mobile guys want to be able to distinguish between crash.log and other logs.

Builds on https://github.com/matrix-org/rageshake/pull/17.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,11 +57,16 @@ logs.)
 * `log`: a log file, with lines separated by newline characters. Multiple log
   files can be included by including several `log` parts.
 
+  If the log is uploaded with a filename `name.ext`, where `name` contains only
+  alphanumerics, `.`, `-` or `_`, and `ext` is one of `log` or `txt`, then the
+  file saved to disk is based on that. Otherwise, a suitable name is
+  constructed.
+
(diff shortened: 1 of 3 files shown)


Kegsay:
It'd be nice to have a test which contains a file name which doesn't meet the regexp requirements to check it does make a sensible name.

Otherwise LGTM.

richvdh:
> It'd be nice to have a test which contains a file name which doesn't meet the regexp requirements to check it does make a sensible name.

https://github.com/matrix-org/rageshake/blob/6892a1a0a5af8c16d7a4e05b830fb442de44a261/src/github.com/matrix-org/rageshake/submit_test.go#L152 cf https://github.com/matrix-org/rageshake/blob/6892a1a0a5af8c16d7a4e05b830fb442de44a261/src/github.com/matrix-org/rageshake/submit_test.go#L206

Admittedly it would be nice if it was a separate test rather than a single "check everything" one, but we can't have everything

Kegsay:
Oh I see, because it doesn't end in `.log` or `.txt`. Right, WFM then.

PR #19, opened 2017-05-10
Unbreak github submits for json rageshakes

richvdh:
Make sure we read the appname from JSON rageshakes, so that issues are correctly filed into github.

(This was affecting riot-web 0.9.9 and earlier; later versions are using the multipart interface; riot-android was special-cased anyway; riot-ios has never used the JSON interface.)

(And this, children, is why we write tests.)


Kegsay:
D'oh. LGTM

PR #20, opened 2017-09-06
Try to avoid setting an empty github issue title

richvdh:
Github complains if the title is empty, so work a bit harder on making sure
that we have a valid title even if the first line of the text is empty.

### DIFF
--- a/src/github.com/matrix-org/rageshake/submit.go
+++ b/src/github.com/matrix-org/rageshake/submit.go
@@ -451,15 +451,16 @@ func (s *submitServer) saveReport(ctx context.Context, p parsedPayload, reportDi
 }
 
 func buildGithubIssueRequest(p parsedPayload, listingURL string) github.IssueRequest {
+	// set the title to the first (non-empty) line of the user's report, if any
 	var title string
-	if p.UserText == "" {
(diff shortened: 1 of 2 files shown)


src/github.com/matrix-org/rageshake/submit.go
```diff
@@ -451,15 +451,16 @@ func (s *submitServer) saveReport(ctx context.Context, p parsedPayload, reportDi
...
 	var title string
-	if p.UserText == "" {
+	trimmedUserText := strings.TrimSpace(p.UserText)
+	if trimmedUserText == "" {
 		title = "Untitled report"
 	} else {
-		// set the title to the first line of the user's report
-		if i := strings.IndexAny(p.UserText, "\r\n"); i < 0 {
-			title = p.UserText
+		if i := strings.IndexAny(trimmedUserText, "\r\n"); i < 0 {
```
erikjohnston: What happens if i == 0? I.e. it starts with `\r\n`? Oh, because we've trimmed it? Fair.

erikjohnston:
lgtm? probably? maybe have a test for empty text?

richvdh:
done

PR #25, opened 2019-04-09
sort lexicographically data fields

### LINKED ISSUE #23: Data fields written in random order
The fields from `p.Data` (such as "device_id", "Version", etc.) are sent unsorted to the details file and GH issue (presumably because we're iterating a Go map which may put them in any order).

It would be better to sort first, so that multiple reports are easier to scan.

gernest:
Fixes #23

Signed-off-by: Geofrey Ernest <geofreyernest@live.com>

### DIFF
--- a/src/github.com/matrix-org/rageshake/submit.go
+++ b/src/github.com/matrix-org/rageshake/submit.go
@@ -81,7 +81,7 @@ type parsedPayload struct {
 	FileErrors []string
 }
 
-func (p parsedPayload) WriteTo(out io.Writer) error {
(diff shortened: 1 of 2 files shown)


gernest:
CI failure has nothing to do with this PR `.travis.yml` is broken

richvdh:
Please could you sign off your contribution as per https://github.com/matrix-org/rageshake/blob/master/CONTRIBUTING.rst#attribution

gernest:
@richvdh thanks for reviewing, please see the top most comment , I added the sign-off line.

Issue #24, opened 2019-04-09
Small improvements for the quality of the project


gernest:
So,  I'm sold on matrix api's and I have been lurking on the matrix-org looking for things I might be useful of (mostly Go)

These are some of improvements I think will make this project easy to work with

-  use `go.mod` to manage dependencies ( meaning vendor dir should go away)
- move source file out of  the `src` directory the project name is already `rageshake` so no need to nest the sources (the layout is confusing for go devs, it is familiar for java devs)

# Why?
I would like to submit more patches , this looks like an interesting tool in the matrix verse and out of habit ( been doing open source go for some time and quality has been very high on  my priorities) this small details matter.

I just need a thumbs up, and I will do all of that and submit a PR in no time. I am asking because I don't want to crash with your philosophy or preference I just want to help so a PR would have been rude(to me) , without asking if its okay to do it.

Thanks for your amazing work.

richvdh:
Yes, these are things we've been doing on some of our other projects (eg https://github.com/matrix-org/matrix-federation-tester/pull/46). We just haven't got around to it here...

PR #26, opened 2019-04-10
move files to the source root

gernest:
see #24 

Signed-off-by: Geofrey Ernest <geofreyernest@live.com>


richvdh:
I've pulled this into #31, along with the other bits necessary to make go modules work. Thanks.

PR #27, opened 2019-04-10
fix travis-ci

gernest:


.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: go
 go:
- - 1.7
+ - 1.12
```
richvdh: Is 1.12 now the earliest version of go that we can build on? If so, we should update the README, but I'd prefer us to maintain compatibility with slightly older versions if possible.

gernest: I am preparing for introducing go modules. 
this should also work with `1.7`  without any issues.   Only  `context` package will limit how far we can go with go versions (can't remember when it was introduced but it is not available in old versions)

#24 is the endgame. So I will be providing a series of patches and will update README accordingly.

gernest:
Gentle ping @richvdh , anything else you want me to do for this?

turt2live:
(I suspect he's busy dealing with Unexpected Other Things, but have gently asked for review for when he returns)

.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: go
 go:
- - 1.7
```
richvdh: (sorry for sitting on this).

If we're preserving compatibility with 1.7, it seems best to leave this as 1.7 so that we can check that. If the project builds ok on 1.7 there's a good chance it will also build ok on 1.12. The reverse is not true.

richvdh:
I'm afraid I remain confused about what a bunch of the changes here were for - certainly they weren't necessary to "fix travis-ci" as the PR subject claimed (see #29).

In any case I think the relevant changes have now been pulled into #29, #30 and #31, so I'm going to go ahead and close this. Thanks for contributing.

PR #28, opened 2019-04-15
Add support for Slack notifications

awesome-manuel:


richvdh:
@awesome-manuel : thanks for contributing this. Unfortunately it's conflicted with our switch to go modules (#31) and the CI wasn't working at the time.

would you be able to update the patch?

awesome-manuel:
@richvdh Updated

submit.go
```diff
@@ -460,36 +462,62 @@ func (s *submitServer) saveReport(ctx context.Context, p parsedPayload, reportDi
...
+	return &resp, nil
+}
 
-	issue, _, err := s.ghClient.Issues.Create(ctx, owner, repo, &issueReq)
-	if err != nil {
-		return nil, err
-	}
+func (s *submitServer) submitGithubIssue(ctx context.Context, p parsedPayload, listingURL string, resp *submitResponse) error {
+	if s.ghClient == nil {
+		log.Println("GH issue submission disabled")
```
richvdh: For the record: generally I prefer an early return to lots of `else` statements (hence `return nil` here), though I'm aware of counterarguments to it.

PR #32, opened 2019-08-27
Add Dockerfile

awesome-manuel:


awesome-manuel:
You can simply mount it into the root of the container: `docker run -it -v ./rageshake.yaml:/rageshake.yaml rageshake:latest`

richvdh:
Ah right. For the record, I needed an absolute path:

```
docker run -it -v `pwd`/rageshake.yaml:/rageshake.yaml rageshake:latest
```

richvdh:
thanks!

PR #33, opened 2020-03-15
Add means to autocomplete issue references to a different repo

t3chguy:
Fixes https://github.com/vector-im/riot-web/issues/9242

Useful for setups where the bug report repo is not the src code repo and relative issue references refer to the bug report repo instead of the src code repo.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2017 Vector Creations Ltd
+Copyright 2017, 2020 Vector Creations Ltd
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -48,7 +48,8 @@ type config struct {
 	// A GitHub personal access token, to create a GitHub issue for each report.
 	GithubToken string `yaml:"github_token"`
 
-	GithubProjectMappings map[string]string `yaml:"github_project_mappings"`
+	GithubProjectMappings       map[string]string `yaml:"github_project_mappings"`
+	AutocompleteProjectMappings map[string]string `yaml:"autocomplete_project_mappings"`
 
 	SlackWebhookURL string `yaml:"slack_webhook_url"`
 }
@@ -112,7 +113,7 @@ func main() {
 	}
 	log.Printf("Using %s/listing as public URI", apiPrefix)
 
-	http.Handle("/api/submit", &submitServer{ghClient, apiPrefix, cfg.GithubProjectMappings, slack})
+	http.Handle("/api/submit", &submitServer{ghClient, apiPrefix, cfg.GithubProjectMappings, cfg.AutocompleteProjectMappings, slack})
 
 	// Make sure bugs directory exists
 	_ = os.Mkdir("bugs", os.ModePerm)
--- a/main.go
+++ b/main.go
@@ -1,5 +1,6 @@
 /*
-Copyright 2017, 2020 Vector Creations Ltd
+Copyright 2017 Vector Creations Ltd
+Copyright 2020 The Matrix.org Foundation C.I.C.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -48,7 +49,10 @@ type config struct {
 	// A GitHub personal access token, to create a GitHub issue for each report.
 	GithubToken string `yaml:"github_token"`
 
-	GithubProjectMappings       map[string]string `yaml:"github_project_mappings"`
+	// Mappings from app name (as submitted in the API) to github repo for issue reporting.
+	GithubProjectMappings map[string]string `yaml:"github_project_mappings"`
+	// Mappings from app name (as submitted in the API) to github repo as to which the issues pertain.
+	// Not needed if the issues are reported to the main repo as github will complete the ambiguous references correctly.
 	AutocompleteProjectMappings map[string]string `yaml:"autocomplete_project_mappings"`
 
 	SlackWebhookURL string `yaml:"slack_webhook_url"`
--- a/rageshake.sample.yaml
+++ b/rageshake.sample.yaml
@@ -17,6 +17,11 @@ github_token: secrettoken
 github_project_mappings:
    my-app: octocat/HelloWorld
 
+# mappings from app name (as submitted in the API) to github repo as to which the issues pertain.
+# not needed if the issues are reported to the main repo as github will complete the ambiguous references correctly.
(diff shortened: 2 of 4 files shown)


submit.go
```diff
@@ -118,6 +119,9 @@ type submitResponse struct {
 	ReportURL string `json:"report_url,omitempty"`
 }
 
+// regex to catch and substitute ambiguous issue references with explicit ones to the actual repo they are in
+var ambiguousIssueRegex = regexp.MustCompile(`[\s,.](#\d+)`)
```
richvdh: * won't this remove the preceding character?
* will it match the first word in the report?
* will it match foo/#33? I don't think it will, but it should. It feels like it would be better to specify the characters which *cannot* precede the reference.
* will it match #33foo? I think it will, but it shouldn't.

t3chguy: > specify the characters which cannot precede the reference.

golang's regexp lib doesn't have negative lookahead due to its guarantee of O(n) efficiency :(

richvdh: that would be lookbehind.

but that doesn't matter, you can use a capturing group and put the captured text back in?

t3chguy: > will it match foo/#33? I don't think it will, but it should. It feels like it would be better to specify the characters which cannot precede the reference.

What would foo/#33 map to? owner/repo/foo/#33 doesn't make much sense imo

t3chguy: > It feels like it would be better to specify the characters which cannot precede the reference.

In this situation I disagree, it can only be preceded by whitespace or brackets (or nothing if anchored at the beginning), negating that would be much longer, no?

richvdh: I'd have thought it would make sense to follow what github does.

t3chguy: Bases what it does based on whether or not a local issue of that ID exists which seems overkill.
foo/#33
test/#123
I think only accepting a `#NNNNN` token on its own, or maybe in brackets should be accepted. The rest should be handled by github UI itself.

t3chguy: @richvdh if you have alternate formats that would like this "autocomplete" applied to them then please specify them here, will add to tests and get regex to accept them.

main.go
```diff
@@ -48,7 +49,11 @@ type config struct {
 	// A GitHub personal access token, to create a GitHub issue for each report.
 	GithubToken string `yaml:"github_token"`
 
+	// Mappings from app name (as submitted in the API) to github repo for issue reporting.
 	GithubProjectMappings map[string]string `yaml:"github_project_mappings"`
+	// Mappings from app name (as submitted in the API) to github repo as to which the issues pertain.
```
richvdh:
```suggestion
	// Mappings from app name (as submitted in the API) to github repo to which the issues pertain.
```

rageshake.sample.yaml
```diff
@@ -17,6 +17,11 @@ github_token: secrettoken
 github_project_mappings:
    my-app: octocat/HelloWorld
 
+# mappings from app name (as submitted in the API) to github repo as to which the issues pertain.
```
richvdh:
```suggestion
# mappings from app name (as submitted in the API) to github repo to which the issues pertain.
```

submit.go
```diff
@@ -118,6 +120,14 @@ type submitResponse struct {
 	ReportURL string `json:"report_url,omitempty"`
 }
 
+// regex to catch and substitute ambiguous issue references with explicit ones to the actual repo they are in
+var ambiguousIssueRegex = regexp.MustCompile(`(^|[([{\s])(#\d+)([^\w]|$)`)
```
richvdh: could you write a comment which describes what this is matching? It's quite hard to grok by eye. (though having stared at it for several minutes, I think I understand it and it looks plausible).

submit.go
richvdh: > Bases what it does based on whether or not a local issue of that ID exists which seems overkill.

I wasn't suggesting that, fwiw. I was just suggesting that you add `/` to the list of characters which can precede an ambiguous reference (so `foo/#33` would map to `foo/owner/project#33`). It doesn't really matter though.

t3chguy: foo/owner/project#33 won't link though so whats the point of translating it?

richvdh: foo/matrix-org/rageshake#33

t3chguy: Yes, but that'd get mapped to foo/owner/project#33 which no longer makes sense/works

submit.go
```diff
@@ -120,11 +120,13 @@ type submitResponse struct {
 	ReportURL string `json:"report_url,omitempty"`
 }
 
-// regex to catch and substitute ambiguous issue references with explicit ones to the actual repo they are in
-var ambiguousIssueRegex = regexp.MustCompile(`(^|[([{\s])(#\d+)([^\w]|$)`)
+// regex to match and substitute ambiguous issue references in the rageshake body text
+// matches a hash followed by digits optionally surrounded by whitespace or some punctuation
```
richvdh, line 124: this isn't really sufficient, I'm afraid. I'm looking for an explanation of how it works without having to stare hard at it and find the go regex docs.

```suggestion
//
// This matches:
//    (^|[([{\s]) - the start of the text, one of '(' or '[', '{', or a whitespace character
//    (?:^|#) - the start of the text, or a #
//    (\d+) - one or more digits
//    ([^\w]|$) - a non-word character (?) or the end of the text.
```

submit_test.go
```diff
@@ -492,6 +492,8 @@ func TestAutocompleteIssueReferences(t *testing.T) {
 		"test (#123) bar":     "test (owner/repo#123) bar",     // brackets
 		"Start #123. Now":     "Start owner/repo#123. Now",     // followed by punctuation
 		"#123foo":             "#123foo",                       // ignore
+		"foo/#123":            "foo/#123",                      // ignore
+		"123":                 "owner/repo#123",                // special case for entire body being a number
```
richvdh, line 496: nb that the regex above also matches "123 foo": the comment here makes me wonder if that is deliberate?

submit.go
```diff
@@ -120,11 +120,13 @@ type submitResponse struct {
...
+// also matches if the input starts with digits where the hash becomes optional
```
richvdh, line 125: I'm a bit surprised that you really want to match things that start with digits... is this is a common pattern?

in any case: imho there comes a point where you should just use a separate `if` condition, rather than an increasingly complex regex...

richvdh:
I'm going to close this, as it's bitrotting and hasn't moved in over a year. If you'd like to pick it up again, I'd be delighted to reopen it!

PR #35, opened 2020-09-02
Add email support

MatMaul:


MatMaul:
Should I make sending logs&screenshot as attachments an option ?

submit.go
```diff
@@ -570,6 +586,41 @@ func buildGithubIssueRequest(p parsedPayload, listingURL string) github.IssueReq
 	}
 }
 
+func (s *submitServer) sendEmail(p parsedPayload, reportDir string) error {
+	if s.cfg.SMTPServer == "" {
+		log.Println("Email submission disabled")
```
richvdh, line 591: feels like this is a bit redundant and will be a bit noisy on most servers?

submit.go
```diff
@@ -570,6 +586,41 @@ func buildGithubIssueRequest(p parsedPayload, listingURL string) github.IssueReq
...
+	} else {
```
richvdh, line 592: an early return here would be nice, to save the big indented block.

richvdh (approved):
lgtm!

submit.go
MatMaul: agreed, but I went for consistency because it does that for GH issues notifs or Slack one. Should I remove them on the way ?

richvdh: yeah, probably.

submit.go
```diff
@@ -470,70 +471,69 @@ func (s *submitServer) saveReport(ctx context.Context, p parsedPayload, reportDi
...
 
+	if err := s.sendEmail(p, reportDir); err != nil {
+		return nil, err
+	}
+
 	return &resp, nil
 }
 
 func (s *submitServer) submitGithubIssue(ctx context.Context, p parsedPayload, listingURL string, resp *submitResponse) error {
-	if s.ghClient == nil {
```
richvdh, line 477: I'm failing to see where we now handle the cases of github/slack/email being disabled? we seem to call `submitGithubIssue` unconditionally so won't it just fail?

richvdh (changes requested):
as above

submit.go
MatMaul: ... Indeed. Fixing that.

rageshake.sample.yaml
```diff
@@ -20,3 +20,17 @@ github_project_mappings:
 # a Slack personal webhook URL (https://api.slack.com/incoming-webhooks), which
 # will be used to post a notification on Slack for each report.
 slack_webhook_url: https://hooks.slack.com/services/TTTTTTT/XXXXXXXXXX/YYYYYYYYYYY
+
+# notification can also be pushed by email.
+# this param controls the target emails
+email_addresses:
+- support@matrix.org
```
richvdh, line 27: I think this is clearer:

```suggestion
  - support@matrix.org
```

richvdh (approved):
lgtm otherwise

PR #36, opened 2021-06-01
Docker will need the config file at toplevel

jthoma:
I was attempting with the docker build, docker-compose up -d and the container were continuously restarting. Checking the logs showed the entry point was restarting with "rageshake.yaml" not found. I just added the docker-compose file also, since I find that easier than remembering the commands.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -installsuffix cgo -ldflag
 FROM scratch
 COPY --from=builder /build/rageshake /rageshake
 COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=builder /build/rageshake.sample.yaml /rageshake.yaml
 WORKDIR /
 EXPOSE 9110
 CMD ["/rageshake"]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -installsuffix cgo -ldflag
 FROM scratch
 COPY --from=builder /build/rageshake /rageshake
 COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=builder /build/rageshake.sample.yaml /rageshake.yaml
```
richvdh, line 16: I'm not convinced this makes much sense. The sample config file isn't ready for use without customisation.

richvdh (changes requested):

Dockerfile
jthoma: Yes I do agree.. but no reference as to that it should be renamed to rageshake.yaml is anywhere in the install guide or so. Hence I just made the customisations and left the file as is.

jthoma: But concise information is given that the customisations should be done in the rageshake.sample.yaml

jthoma (commented):
Made some changes, such that the docker can expect a customized file. Complimentary addition is done to the README.md also.

richvdh (changes requested):
This doesn't appear to work:

```
$ docker build .
...
Step 13/16 : COPY --from=builder /build/rageshake.yaml /rageshake.yaml
COPY failed: stat build/rageshake.yaml: file does not exist
```

I don't really think it makes sense to include a `rageshake.yaml` file inside the docker image, since everybody will need their own version. 

As noted in #32, the intention is that you supply your own `rageshake.yaml` with a `-v` option. If you'd like to update the documentation to make that clear, that would be welcome.

richvdh:
I'm going to close this, as it seems to be abandoned.

PR #37, opened 2021-08-06
Add support for creating GitLab issues

tulir:
This adds support for creating GitLab issues using https://github.com/xanzy/go-gitlab.

I also updated the report body to have two spaces before newlines so it would work better as markdown. I'd assume GitHub doesn't care either way, but GitLab requires those.

Signed-off-by: Tulir Asokan &lt;tulir@maunium.net&gt;


go.mod
```diff
@@ -1,17 +1,11 @@
 module github.com/matrix-org/rageshake
 
+go 1.16
```
richvdh, line 3: the readme claims "rageshake requires Go version 1.11 or later."

I have no particular desire to support ancient Go versions, but we better update the documentation and the CI if we want to update it.

submit.go
```diff
@@ -541,7 +570,7 @@ func buildReportTitle(p parsedPayload) string {
 	return trimmedUserText
 }
 
-func buildReportBody(p parsedPayload, quoteChar string) *bytes.Buffer {
+func buildReportBody(p parsedPayload, newline, quoteChar string) *bytes.Buffer {
```
richvdh, line 573: this is getting a bit confusing. A comment explaining what exactly the parameters do would be helpful.

richvdh: I'm also not entirely convinced this needs to be different between email and github/gitlab?

tulir: Email isn't markdown and I didn't want to accidentally break something with extra spaces at the end of each line, but I guess it might be fine to add the spaces for emails too. (GitHub is markdown, but doesn't seem to require the two spaces for newlines. GitLab does require spaces to make a newline)

go.mod
tulir: Updated the readme and go.mod to say 1.15 and made matrix-org/pipelines#171 to update the CI.

rageshake.sample.yaml
```diff
@@ -17,6 +17,23 @@ github_token: secrettoken
...
+gitlab_url: https://gitlab.com
+
+# mappings from app name (as submitted in the API) to the GitLab Project ID (not name!) for issue reporting.
+gitlab_project_mappings:
+  my-app: 12345
+# mappings from app name to a list of GitLab label names for issue reporting.
+gitlab_project_labels:
+  my-app:
+    - client::my-app
+# whether GitLab issues should be created as confidential issues
```
richvdh, line 34:
```suggestion
# whether GitLab issues should be created as confidential issues. Defaults to false.
```

richvdh (commented):
lgtm otherwise

richvdh:
Thanks @tulir!

PR #38, opened 2021-12-06
Fix support for authenticated SMTP with a custom port

awesome-michael:
Sending emails fails if the port is used in smtp.PlainAuth. Since no timeout is configured this error blocks the complete application.
I separated host and port of the SMTP server in the config to prevent this.

### DIFF
--- a/changelog.d/38.bugfix
+++ b/changelog.d/38.bugfix
@@ -1,0 +1,1 @@
+Fix email support to support mail setups with different ports than 25.
--- a/changelog.d/38.bugfix
+++ b/changelog.d/38.bugfix
@@ -1,1 +1,1 @@
-Fix email support to support mail setups with different ports than 25.
+Fix email support to support authenticated mail with a custom port.
--- a/changelog.d/38.misc
+++ b/changelog.d/38.misc
@@ -1,1 +1,0 @@
-Fix email support.
--- a/changelog.d/38.misc
+++ b/changelog.d/38.misc
@@ -1,0 +1,1 @@
+Fix email support.
--- a/submit.go
+++ b/submit.go
@@ -666,7 +666,7 @@ func (s *submitServer) sendEmail(p parsedPayload, reportDir string) error {
 	if s.cfg.SMTPPassword != "" || s.cfg.SMTPUsername != "" {
 		auth = smtp.PlainAuth("", s.cfg.SMTPUsername, s.cfg.SMTPPassword, s.cfg.SMTPServer)
 	}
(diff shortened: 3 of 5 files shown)


changelog.d/38.misc
```diff
@@ -0,0 +1 @@
+Fix email support.
```
richvdh, line 1: please could you expand on this a bit? this implies it was completely broken before, which I don't think is the case?

(also, this should be a `.bugfix` file)

submit.go
```diff
@@ -666,7 +666,7 @@ func (s *submitServer) sendEmail(p parsedPayload, reportDir string) error {
 	if s.cfg.SMTPPassword != "" || s.cfg.SMTPUsername != "" {
 		auth = smtp.PlainAuth("", s.cfg.SMTPUsername, s.cfg.SMTPPassword, s.cfg.SMTPServer)
 	}
-	err := e.Send(s.cfg.SMTPServer, auth)
+	err := e.Send(s.cfg.SMTPServer+":"+s.cfg.SMTPServerPort, auth)
```
richvdh, line 669: it might be preferable to use `net.SplitHostPort` (https://pkg.go.dev/net#SplitHostPort) to split the setting into server and port parts, to avoid breaking backwards compatibility.

richvdh (changes requested):
Thanks for the contribution!

changelog.d/38.misc
awesome-michael: will rename the file and try to expand the description. Although mailing was not working for me at all. But it may depend on the setup

changelog.d/38.bugfix
```diff
@@ -0,0 +1 @@
+Fix email support to support mail setups with different ports than 25.
```
michaelkaye, line 1:
```suggestion
Fix email support to support authenticated mail with a custom port.
```

michaelkaye (approved):
This looks ok by me, would suggest this change to the changelog as we explicitly mention that it's authentication where there are issues with the current code.

changelog.d/38.bugfix
awesome-michael: ack

richvdh (approved):
thanks!

richvdh:
@awesome-michael please could you add a "signed-off-by" to this PR, per https://github.com/matrix-org/rageshake/blob/master/CONTRIBUTING.rst#sign-off?

awesome-michael:
the last commit is signed-off

PR #39, opened 2021-12-14
Add random string to report directories to prevent time-based collisions

duxovni:
Fixes a bug I ran into while testing matrix-org/matrix-react-sdk#7307, which causes receiving and sending clients to submit rageshakes near-simultaneously. If the timestamps collide, only one client's rageshake succeeds, and the other one fails with a 500.

### DIFF
--- a/changelog.d/39.bugfix
+++ b/changelog.d/39.bugfix
@@ -1,0 +1,1 @@
+Prevent timestamp collisions when reports are submitted within 1 second of each other.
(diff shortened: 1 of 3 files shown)


t3chguy:
Did this skip review?

duxovni:
I guess? I'm not super clear on what the review process for this repo is, sorry about that!

t3chguy:
For future reference its a good idea to look at other open or closed PRs to see what lifecycle they went through

Issue #40, opened 2022-01-18
Rageshake server does not accept "log.gz" as a valid file suffix", as used by element-android


michaelkaye:
**Describe the bug**

element-android submits rageshakes with gzipped file parts, eg:

https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/rageshake/BugReporter.kt#L700 creates gzipped files named with a .gz extension

https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/rageshake/BugReporter.kt#L299 creates Parts with the filename set to the name

Rageshake server rejects the name and overwrites with a good one as it doesn't match the regular expression here:

https://github.com/matrix-org/rageshake/blob/master/submit.go#L427

It then replaces it with a logs-0000.gz style extension.

This makes it hard to handle crash data, logcat data and log files separately.

So rather than getting the following files:
```
logcat.log.gz
logs-1.log.gz
logs-2.log.gz
logs-3.log.gz
logs-4.log.gz
logs-5.log.gz
logs-6.log.gz
logs-7.log.gz
```

We get the following files:
```
logs-0000.txt.gz (this is the logcat one)
logs-0001.txt.gz
logs-0002.txt.gz
logs-0003.txt.gz
logs-0004.txt.gz
logs-0005.txt.gz
logs-0006.txt.gz
logs-0007.txt.gz
```

Unsure whether to fix in the rageshake server (and accept *.gz as a valid extension) or if we should fix in element-android going forward and accept currently deployed apps'  log submissions will be made into generic logs-0000.txt.gz files.

michaelkaye:
This is also true of element-web, which submits logs (i believe) with a name similar to `instance-123152` - however it's less of a problem there as all logs are of the same type; and it submits them in reverse chronological order, so it's easier to manage.

michaelkaye:
This is now deployed on the matrix.org server, logs are now named correctly when uploaded.

bmarty:
Thanks a lot for this fix @michaelkaye ! I was wondering since a long time why we had this diff in the files name.
I confirm that we now have better filenames on the rageshakes:
<img width="140" alt="image" src="https://user-images.githubusercontent.com/3940906/151204464-9584144a-058e-40d9-83f1-8db763edeee8.png">

PR #41, opened 2022-01-18
Allow files ending "txt.gz" to be uploaded as a log file.

michaelkaye:
Addresses #40 by allowing the log.gz extension to be used.

### DIFF
--- a/submit.go
+++ b/submit.go
@@ -424,7 +424,7 @@ func saveFormPart(leafName string, reader io.Reader, reportDir string) (string, 
 
 // we require a sensible extension, and don't allow the filename to start with
 // '.'
-var logRegexp = regexp.MustCompile(`^[a-zA-Z0-9_-][a-zA-Z0-9_.-]*\.(log|txt)$`)
+var logRegexp = regexp.MustCompile(`^[a-zA-Z0-9_-][a-zA-Z0-9_.-]*\.(log|txt)(\.gz)?$`)
 
 // saveLogPart saves a log upload to the report directory.
 //
--- a/submit_test.go
+++ b/submit_test.go
@@ -160,6 +160,7 @@ func TestMultipartUpload(t *testing.T) {
 
 	// check file uploaded correctly
 	checkUploadedFile(t, reportDir, "passwd.txt", false, "bibblybobbly")
+	checkUploadedFile(t, reportDir, "crash.log.gz.gz", true, "test\n")
 }
 
 func multipartBody() (body string) {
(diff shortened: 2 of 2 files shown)


submit.go
```diff
@@ -438,7 +438,11 @@ func saveLogPart(logNum int, filename string, reader io.Reader, reportDir string
 	// Either way, we need to append .gz, because we're compressing it.
 	var leafName string
 	if logRegexp.MatchString(filename) {
-		leafName = filename + ".gz"
+		if strings.HasSuffix(filename, ".gz") {
+			leafName = filename
+		} else {
+			leafName = filename + ".gz"
+		}
```
richvdh, line 445: won't we still compress the log file even if it's already gzipped? maybe that's a silly thing to do, but it should be called `foo.log.gz.gz` if we're doing it.

michaelkaye: We unzip the file as part of the `compressed-log` form field handling, then in this code we gzip it again. I think the only place this would go wrong is if someone uploaded a gzip file to the uncompressed `log` form field (already an issue, you're meant to upload to compressed-log) - if you name it with a gz then you end up with only one gz, not two; if you don't name it with a gz you also get one.

The code would also do the wrong thing, if you upload a file for instance "foo.gz" in the `log` form-field - since the filename doesn't match the expression, it's overwritten with the fixed "logs-0000.log.gz` filename - so that too would get the compression status out of sync with the filename.

I looked at doing something in the `compressed-log` side before it came in to this method, but trying to add something into gzip.NewPart handling to update a Header if it happens to be called 'filename' and ends 'gz' seemed more work.

richvdh: > We unzip the file as part of the compressed-log form field handling, then in this code we gzip it again

right, fair enough. In that case, could I ask you to add a comment to explain that this is what's going in?

submit.go
```diff
@@ -438,7 +438,11 @@ func saveLogPart(logNum int, filename string, reader io.Reader, reportDir string
 	// Either way, we need to append .gz, because we're compressing it.
```
richvdh, line 438: ... maybe. 

Could you update this pls?

submit.go
richvdh, line 445: personally I would find this more intuitive:

```suggestion
		leafName = filename               
		if !strings.HasSuffix(filename, ".gz") {
			leafName += "gz"
		}
```

... ymmv though.

michaelkaye: I've updated the comment here.

richvdh (approved):

PR #42, opened 2022-01-18
Update go version to 1.16

michaelkaye:
The linting scripts don't run on 1.15 any more, so builds are not
running against 1.15 any more.

Technically it will still compile and run against go 1.15 but if we
can't verify it in CI, it's not known to be true in future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Web service which collects and serves bug reports.
 
-rageshake requires Go version 1.15 or later.
+rageshake requires Go version 1.16 or later.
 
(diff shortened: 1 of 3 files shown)


richvdh:
presumably you plan to update the deployment somehow?

michaelkaye:
So current deployment for the element instance is from the released executable https://github.com/matrix-org/rageshake/releases - I'm still working out how that is built, but yeah, will be deploying this as part of a few related updates.

richvdh (approved):

PR #47, opened 2022-01-20
Build and push multi-arch Docker images on ghcr.io

sandhose:
This does two things:
 - enhance the Dockerfile to be multi-arch capable (also allows it to run as non-root)
 - build and push a Docker image on ghcr.io/matrix-org/rageshake via GitHub Actions

There are a few important things to note about the changes in the Dockerfile:
 - the runtime image is based on an image [from the distroless project](https://github.com/GoogleContainerTools/distroless)
 - this allows for the container to run as non-root (there is a `/etc/passwd` entry for that)
 - the image can be cross-compiled (e.g. building the amd64 image from an arm64 host and vice-versa) with no qemu emulation involved. This however requires a buildkit-enabled Docker (which is the default on Docker Desktop), e.g. `export DOCKER_BUILDKIT=1` on Linux
 - I tested the image both on my M1 MacBook (arm64) and my x86_64 desktop
 - I switched the `/rageshake` binary to be the entrypoint instead of the command. This allows to run the image like `docker run ghcr.io/matrix-org/rageshake -help` instead of `docker run ghcr.io/matrix-org/rageshake /rageshake -help`

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -17,6 +17,9 @@ jobs:
       contents: read
 
     steps:
+      - name: Checkout the code
+        uses: actions/checkout@v2
+
       - name: Docker meta
         id: meta
         uses: docker/metadata-action@v3
@@ -25,11 +28,27 @@ jobs:
           bake-target: docker-metadata-action
           tags: |
             type=ref,event=branch
-            type=ref,event=pr
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
             type=sha
+
+      - name: Docker meta (debug variant)
+        id: meta-debug
+        uses: docker/metadata-action@v3
+        with:
+          images: "${{ env.IMAGE }}"
+          bake-target: docker-metadata-action-debug
+          tags: |
+            type=ref,event=branch,suffix=-debug
+            type=semver,pattern={{version}},suffix=-debug
+            type=semver,pattern={{major}}.{{minor}},suffix=-debug
+            type=semver,pattern={{major}},suffix=-debug
+            type=sha,suffix=-debug
+
+      - name: Merge buildx bake files
+        run: |
+            jq -s '.[0] * .[1]' ${{ steps.meta.outputs.bake-file }} ${{ steps.meta-debug.outputs.bake-file }} > docker-bake.override.json
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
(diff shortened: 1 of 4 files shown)


sandhose:
It happened: I needed to pop a shell within the container to debug something, but the image does not include one.
I added a `debug` variant of the image (based on the `debug` variant of the `distroless` image), and to avoid building the image from scratch twice, I switched to Docker Buildx bake files, which allows specifying multiple targets at once, building only one build graph, which in the end avoids compiling the Go binary twice for both variants.

I have a similar setup over at https://github.com/matrix-org/matrix-authentication-service

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,84 @@
...
+        uses: docker/metadata-action@v3
+        with:
+          images: "${{ env.IMAGE }}"
+          bake-target: docker-metadata-action
+          tags: |
+            type=ref,event=branch
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+            type=sha
```
richvdh, line 34: I'm struggling to get my head round this a bit. is the first match in the list used? or what?

Dockerfile
```diff
@@ -1,18 +1,27 @@
...
+RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o rageshake
+
+## Runtime stage, debug variant ##
+FROM --platform=${TARGETPLATFORM} gcr.io/distroless/static-debian${DEBIAN_VERSION}:debug-nonroot AS debug
+COPY --from=builder /build/rageshake /rageshake
+WORKDIR /
+EXPOSE 9110
+ENTRYPOINT ["/rageshake"]
 
-FROM scratch
```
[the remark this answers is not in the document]
sandhose: The main benefit here is having proper entries in `/etc/passwd` and `/etc/group` for a `nobody` user/group which allows running as non-root.
Other than that, it's a nice base image with basically tzdata, ca-certificates, and a few things one can expect from a standard env in `/etc`

Dockerfile
```diff
@@ -1,18 +1,27 @@
...
-COPY go.sum .
+COPY go.mod go.sum ./
 RUN go mod download
 
 COPY . .
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -installsuffix cgo -ldflags '-extldflags "-static"' -o rageshake
+RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o rageshake
+
+## Runtime stage, debug variant ##
+FROM --platform=${TARGETPLATFORM} gcr.io/distroless/static-debian${DEBIAN_VERSION}:debug-nonroot AS debug
```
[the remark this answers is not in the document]
sandhose: Yep, it's the default. It's just to make it clear that some stages run on the host platform while others are on the target platform.

.github/workflows/docker.yaml
sandhose: Those are the rules for generating the list of tags. It's documented here: https://github.com/docker/metadata-action#tags-input

`type=ref,event=branch`: a push to a branch named `foo` creates a tag named `foo`
`type=semver,pattern={{version}}`: a `v1.2.3` git tag creates a `1.2.3` docker tag. I think it also sets the `latest` tag?
`type=semver,pattern={{major}}.{{minor}}`: a `v1.2.3` git tag creates a `1.2` docker tag
`type=semver,pattern={{major}}`: a `v1.2.3` git tag creates a `1` docker tag
`type=sha`: any commit generates a `sha-aabbccd`, where `aabbccd` are the 7 first digits of the commit hash

The rules below are the same, but with a `-debug` suffix (e.g. `:1.2.3-debug`)

sandhose: So the latest push on this PR branch (`sandhose/docker-improvements` / ab3c3d8e460e3d34e4bd53d813b8990abc260516) pushed the following tags:

- `sha-ab3c3d8-debug` and `sandhose-docker-improvements-debug` for the debug variant
- `sha-ab3c3d8` and `sandhose-docker-improvements` for the regular image

See https://github.com/matrix-org/rageshake/pkgs/container/rageshake

richvdh: oh so they are *all* applied. Sorry, I was having a bit of a brainfail.

richvdh (approved):
:shrug: looks plausible to me

PR #50, opened 2022-01-31
Initial attempt at a generic webhook for reports.

michaelkaye:
Background submission to a notification server that doesn't block rageshake submission.

Timeout requests after 300s (to provide some wiggle room for servers that choose to take action inline)

Add docs on the format of the endpoint (though much of it is pass-through so there is no fixed schema in most cases)

### DIFF
--- a/docs/generic_webhook.md
+++ b/docs/generic_webhook.md
@@ -1,0 +1,38 @@
+## Generic webhook request
+
+If the configuration option `generic_webhook_url` is set, then a synchronous request to
+the endpoint will be sent after the incoming request is parsed and the files are uploaded.
+
+The webhook is designed for notification or other tracking services, and does not contain
+the original log files uploaded.
+
+(If you want the original log files, we suggest to implement the rageshake interface itself).
+
+A sample JSON body is as follows:
+
+```
+{
+  'user_text': 'test\r\n\r\nIssue: No issue link given',
+  'app': 'element-web',
+  'data': {
+    'User-Agent': 'Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:96.0) Gecko/20100101 Firefox/96.0',
+    'Version': '0f15ba34cdf5-react-0f15ba34cdf5-js-0f15ba34cdf5',
+    ...
+    'user_id': '@michaelgoesforawalk:matrix.org'},
+  'labels': None,
+  'logs': [
+    'logs-0000.log.gz',
+    'logs-0001.log.gz',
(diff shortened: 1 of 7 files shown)


michaelkaye:
I'm unsure what the WriteTo error in linting is about at the moment; I haven't adjusted that method or changed calls to it (as far as I'm aware).

michaelkaye:
> I'm unsure what the WriteTo error in linting is about at the moment; I haven't adjusted that method or changed calls to it (as far as I'm aware).

Ah, ignore me (ish). Locally, scripts/lint.sh is reporting:

```
# github.com/matrix-org/rageshake
./submit.go:99:24: method WriteTo(out io.Writer) should have signature WriteTo(io.Writer) (int64, error)
```
From the go vet incantation

But the actual error in CI is coming from gocyclo complaining that main.go is at 13 now, not sure why go vet isn't reporting/running in CI the same as locally.

docs/generic_webhook.md
```diff
@@ -0,0 +1,38 @@
+## Generic webhook request
+
+If the configuration option `generic_webhook_url` is set, then a synchronous request to
```
richvdh, line 3: synchronous?

main.go
```diff
@@ -71,6 +71,8 @@ type config struct {
 	SMTPUsername string `yaml:"smtp_username"`
 
 	SMTPPassword string `yaml:"smtp_password"`
+
+	GenericWebhookURL string `yaml:"generic_webhook_url"`
```
richvdh, line 75: this should probably be a list of urls, rather than a single one.

richvdh (commented):
seems generally plausible

docs/generic_webhook.md
michaelkaye: It started synchronous, but became async because we don't want to expose this to the end user.

submit.go
```diff
@@ -521,24 +521,26 @@ func (s *submitServer) submitGenericWebhook(p parsedPayload, listingURL string,
...
-	if err != nil {
-		return err
+	for _, url := range s.cfg.GenericWebhookURLs {
+		// Enrich the parsedPayload with a reportURL and listingURL, to convert a single struct
+		// to JSON easily
+		genericHookPayload := genericWebhookPayload{
+			parsedPayload: p,
+			ReportURL: reportURL,
+			ListingURL: listingURL,
+		}
```
richvdh, line 531: hoist outside the loop, since it's independent?

richvdh (approved):
lgtm otherwise

PR #52, opened 2022-02-08
Allow upload of files with a .json postfix.

michaelkaye:
The limit was there to make arbitrary file download hard, json files are mostly treated the same as text files so this should be fine.

### DIFF
--- a/changelog.d/52.feature
+++ b/changelog.d/52.feature
@@ -1,0 +1,1 @@
+Allow upload of Files with a .json postfix.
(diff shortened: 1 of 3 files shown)


richvdh (approved):
consider updating the mime type mapping at https://github.com/matrix-org/rageshake/blob/53e8947cb99f4f363d6ef3fa574537bbd3851798/logserver.go#L107?

LGTM otherwise.

PR #53, opened 2022-02-09
Serve a tarball containing the contents of a given directory.

michaelkaye:
This will make it easier to get all logs for a given bug; preventing users
needing to run scripts to download all files.

 - we cannot make the link exist in the directory listing as there are scripts
   that automate downloads which would pick this up.

 - Unsure if "?format=tar.gz" is the right option to enable this; I couldn't think
   of something easy to do but hard to not get picked up by existing automation, and
   wouldn't conflict with existing filenames. Suggestions welcome.

### DIFF
--- a/logserver.go
+++ b/logserver.go
@@ -70,7 +70,6 @@ func (f *logServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 }
 
 func serveFile(w http.ResponseWriter, r *http.Request, path string) {
-
 	d, err := os.Stat(path)
 	if err != nil {
 		msg, code := toHTTPError(err)
@@ -88,6 +87,8 @@ func serveFile(w http.ResponseWriter, r *http.Request, path string) {
 			log.Println("Serving tarball of", path)
 			err := serveTarball(w, r, path)
 			if err != nil {
+				msg, code := toHTTPError(err)
+				http.Error(w, msg, code)
 				log.Println("Error", err)
 			}
 			return
@@ -142,12 +143,20 @@ func extensionToMimeType(path string) string {
 // 
 // The resultant tarball will contain a single directory containing all the files
 // so it can unpack cleanly without overwriting other files.
+//
+// Errors are only returned if generated before the tarball has started being
+// written to the ResponseWriter
 func serveTarball(w http.ResponseWriter, r *http.Request, dir string) error {
 	directory, err := os.Open(dir)
 	if err != nil {
 		return err
 	}
-	// "disposition filename"
+
+	// Creates a "disposition filename"
+	// Take a URL.path like `/2022-01-10/184843-BZZXEGYH/`
+	// and removes leading and trailing `/` and replaces internal `/` with `_`
+	// to form a suitable filename for use in the content-disposition header
+	// dfilename would turn into `2022-01-10_184843-BZZXEGYH`
 	dfilename := strings.Trim(r.URL.Path,"/")
 	dfilename = strings.Replace(dfilename, "/","_",-1)
 
@@ -157,7 +166,7 @@ func serveTarball(w http.ResponseWriter, r *http.Request, dir string) error {
 	w.Header().Set("Content-Type", "application/gzip")
 	w.Header().Set("Content-Disposition", "attachment; filename=" + dfilename + ".tar.gz")
 
-	filenames, err := directory.Readdirnames(-1)
+	files, err := directory.Readdir(-1)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


logserver.go
```diff
@@ -69,6 +70,7 @@ func (f *logServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 }
 
 func serveFile(w http.ResponseWriter, r *http.Request, path string) {
+
```
richvdh, line 73: spurious?

logserver.go
```diff
@@ -79,9 +81,18 @@ func serveFile(w http.ResponseWriter, r *http.Request, path string) {
...
-	// if it's a directory, serve a listing
+	// if it's a directory, serve a listing or a tarball
 	if d.IsDir() {
-		log.Println("Serving", path)
+		format, _ := r.URL.Query()["format"]
+		if len(format) == 1 && format[0] == "tar.gz" {
+			log.Println("Serving tarball of", path)
+			err := serveTarball(w, r, path)
+			if err != nil {
+				log.Println("Error", err)
```
richvdh, line 91: think you need to call [`http.Error()`](https://pkg.go.dev/net/http#Error) in this case to make sure we serve a response of some sort if there is an error before we serve anything.

(also, `serveTarball` should probably not return an error if an error occurs once streaming has started.)

logserver.go
```diff
@@ -79,9 +81,18 @@ func serveFile(w http.ResponseWriter, r *http.Request, path string) {
...
+			}
+			return
+		}
+		log.Println("Serving directory listing of", path)
 		http.ServeFile(w, r, path)
 		return
```
richvdh, line 97: suggest moving all this out into a separate function

logserver.go
```diff
@@ -125,6 +136,78 @@ func extensionToMimeType(path string) string {
...
+// The resultant tarball will contain a single directory containing all the files
+// so it can unpack cleanly without overwriting other files.
+func serveTarball(w http.ResponseWriter, r *http.Request, dir string) error {
+	directory, err := os.Open(dir)
+	if err != nil {
+		return err
+	}
+	// "disposition filename"
+	dfilename := strings.Trim(r.URL.Path,"/")
+	dfilename = strings.Replace(dfilename, "/","_",-1)
```
richvdh, line 152: so this replaces `/api/listing/2022-02-10/184843-BZZXEGYH` with `/api/listing/2022-02-10_184843-BZZXEGYH` ? is that... useful?

richvdh: hrm, I guess that's not what it's doing. a clearer comment, with an example, would be helpful.

richvdh (changes requested):
looks generally sensible.

Suggestions:
 * document this at https://github.com/matrix-org/rageshake#get-apilisting
 * add a link to the tarball from the GH issue, so that there is a direct link for new rageshakes.

logserver.go
michaelkaye: The path is just the 2022-02-10 onwards; i believe the http.Handler strips the leading /api/listing for us when we route requests here.

logserver.go
```diff
@@ -79,9 +81,18 @@ func serveFile(w http.ResponseWriter, r *http.Request, path string) {
 	// for anti-XSS belt-and-braces, set a very restrictive CSP
 	w.Header().Set("Content-Security-Policy", "default-src: none")
 
-	// if it's a directory, serve a listing
+	// if it's a directory, serve a listing or a tarball
 	if d.IsDir() {
-		log.Println("Serving", path)
+		format, _ := r.URL.Query()["format"]
+		if len(format) == 1 && format[0] == "tar.gz" {
+			log.Println("Serving tarball of", path)
```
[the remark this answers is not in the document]
michaelkaye: Fixed by allowing a download of any directory, but only including files in that level of the directory - not permitting directory traversal at all.

michaelkaye:
> Another suggestion: use a completely different endpoint (eg `/api/archive/...`). YMMV.

I thought about this; my logic is that if you are looking at a directory listing, it's easier to add a ?format=tar.gz on the end than change endpoint.

michaelkaye:
Oh, and explicitly not adding the tarball to the github issues *for now*, i want to get the ability to download working in production before we start scattering links to it (in case we do want to change it to /archive or whatever)

logserver.go
```diff
@@ -83,18 +83,7 @@ func serveFile(w http.ResponseWriter, r *http.Request, path string) {
...
-			if err != nil {
-				log.Println("Error", err)
-			}
-			return
-		}
-		log.Println("Serving directory listing of", path)
-		http.ServeFile(w, r, path)
-		return
+		serveDirectory(w, r, path)
 	}
```
richvdh, line 87: this needs a `return` ? (or you need to guard the rest of it with an `else`)

logserver.go
```diff
@@ -136,18 +125,44 @@ func extensionToMimeType(path string) string {
...
+		if err != nil {
+			msg, code := toHTTPError(err)
+			http.Error(w, msg, code)
+			log.Println("Error", err)
+		}
+		return
+	}
+	log.Println("Serving directory listing of", path)
+	http.ServeFile(w, r, path)
+	return
```
richvdh, line 143: this `return` is redundant

richvdh (approved):
lgtm otherwise

PR #54, opened 2022-04-06
Persist prefix as a unique id in rageshakes

michaelkaye:
Rageshakes are too numerous for just uploading to github any more.

The generic webhook (and other) mechanism are able to track a greater volume of requests than simply uploading to github; however this means that we are going to be unable to use the issue ID generated by the github upload to track the rageshake.

We do create a unique ID for the on-disk rageshake, we expose this prefix as an ID in the objects we send onwards via the generic webhook API, which can then be used in external systems to uniquely identify this rageshake.

Technically this isn't any information that isn't in the ListingURL already, but i feel having a separate parameter storing the unique ID (such that it can evolve separately from the listingURL) is the right way to go.

### DIFF
--- a/submit.go
+++ b/submit.go
@@ -58,7 +58,7 @@ type submitServer struct {
 	slack *slackClient
 
 	genericWebhookClient *http.Client
-	cfg *config
+	cfg                  *config
 }
 
 // the type of payload which can be uploaded as JSON to the submit endpoint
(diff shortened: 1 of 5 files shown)


submit.go
```diff
@@ -77,15 +77,15 @@ type jsonLogEntry struct {
...
 type genericWebhookPayload struct {
 	parsedPayload
-	ReportURL string             `json:"report_url"`
-	ListingURL string            `json:"listing_url"`
+	ReportURL  string `json:"report_url"`
+	ListingURL string `json:"listing_url"`
 }
 
 // the payload after parsing
 type parsedPayload struct {
```
richvdh, line 87: it might be good to rename this type, and/or update the comment, since it's not longer really limited to the payload of the request.

submit.go
```diff
@@ -77,15 +77,15 @@ type jsonLogEntry struct {
...
+	ID         string            `json:"id"`
```
richvdh, line 88: a comment here describing the function of the field might be nice.

michaelkaye: Have tried to comment on all fields now.

richvdh (approved):
lgtm!

PR #55, opened 2022-04-25
Create additional labels when detecting Dendrite or Conduit homeserver

neilalexander:
This should make it easier to spot rageshakes related to users running these homeservers. Currently the `server_version` field is only sent by Element Android though so will need to do something about iOS still.

Some of the formatting is due to running `goimports` on `submit.go` so sorry about the noise with that.

### DIFF
--- a/changelog.d/55.feature
+++ b/changelog.d/55.feature
@@ -1,0 +1,1 @@
+Add Dendrite and Conduit labels when detected in the `server_version` field of the rageshake payload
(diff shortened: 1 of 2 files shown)


t3chguy:
> Currently the server_version field is only sent by Element Android though so will need to do something about iOS still.

Keep in mind clients have no CS API to ask the server for its version, that's on the federation API only https://spec.matrix.org/v1.2/server-server-api/#get_matrixfederationv1version so is unlikely to get implemented across the board as it'll be brittle.

neilalexander:
> Keep in mind clients have no CS API to ask the server for its version, that's on the federation API only https://spec.matrix.org/v1.2/server-server-api/#get_matrixfederationv1version 

Bah, of course. That's annoying.

richvdh:
yeah, I'm a bit :-1: on this change. It feels like it's going to be fragile.

Issue #59, opened 2022-08-26
Rageshake should support defining a retention period


benbz:
**Is your feature request related to a problem? Please describe.**
For the purpose of not holding onto data longer than needed (are Rageshakes older than 1 year still going to contain useful and actionable information?) and for storage management reasons, Rageshake should cleanup rageshakes older than a configurable period of time

**Describe the solution you'd like**
A configuration option in Rageshake defining a retention period and Rageshake having a process to safely remove rageshakes outside of this retention period

**Describe alternatives you've considered**
A cronjob that does this outside of Rageshake. Without knowing more about Rageshake's internals I can't say whether or not this is safe. Regardless all deployments of Rageshake could benefit from the retention period so doing it in Rageshake gives it to all deployments with minimal extra work

**Additional context**
n/a

michaelkaye:
Additionally, we probably want to search for all items that have certain properties recorded in their details.log.gz, for example the application or the version of the code - i can imagine that unreleased code rageshakes have a lower value over time compared to released code - this is probably a infrequent cross-item search use-case that  `grep` on a filesystem would be great for  (vs the object storage in #60).

(and yes, we could additionally implement a database to store the contents of the details.log.gz and be able to search that for objects to remove instead)

michaelkaye:
So while rageshakes itself doesn't provide a retention period; the provided script can easily run as a cronjob to clean up behind the rageshakes server (as we only write in day-by-day folders).

PR #61, opened 2022-09-09
Cleanup script to search and remove rageshakes from applications based on a time

michaelkaye:
This script as a cronjob could be used to tidy up old rageshakes and implement #59


michaelkaye:
Have tried this in dry run mode on the server. and it is selecting the right rageshakes to delete.

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+                elif data[0] == "#":
+                    # comment, ignore
+                    pass
+                elif data[0] == "@":
+                    # mxid
+                    mxids_to_exclude.append(data)
+                else:
+                    raise Exception(
+                        f"Unable to parse --exclude-mxids file on line {lineno + 1}: {data}"
+                    )
```
richvdh, line 188: would generally suggest a `print(..., file=sys.stderr)` rather than a raise: a stacktrace is not helpful to the user.

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+
+    cleanup = Cleanup(
+        application_limits, days_to_check, args.dry_run, args.path, mxids_to_exclude
+    )
+
+    cleanup.cleanup()
+    print(
+        f"I Deleted {cleanup.deleted} of {cleanup.checked} rageshakes. "
```
richvdh, line 196:
```suggestion
        f"I Deleted {cleanup.deleted} of {cleanup.checked} rageshakes, "
```

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+
+class Cleanup(object):
+    def __init__(
+        self,
+        limits: Dict[str, int],
+        days_to_check: List[int],
+        dry_run: bool,
+        root_path: str,
+        mxids_to_exclude: List[str],
+    ):
```
richvdh, line 26: This probably needs a docstring, if only to document what the parameters mean.

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+# ./cleanup.py --dry-run --path /home/rageshakes/store --max-days 100 element-auto-uisi:90
+#
+# No dependencies required beyond a modern python3.
+
+
+class Cleanup(object):
+    def __init__(
+        self,
+        limits: Dict[str, int],
+        days_to_check: List[int],
```
richvdh, line 22: You're calling this with an `Iterable`, so this is a typing violation.

```suggestion
        days_to_check: Iterable[int],
```

Generally it might be a good idea to run the whole thing through `mypy`.

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+        dry_run: bool,
+        root_path: str,
+        mxids_to_exclude: List[str],
+    ):
+        self.dry_run = dry_run
+        self.days_to_check = days_to_check
+        self.limits = limits
+        self.root_path = root_path
```
richvdh, line 30: these seem to be internal properties, so would suggest giving them a `_` prefix to reflect that. Ditto `mxids_to_exclude`.

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+        self.deleted = 0
+        self.checked = 0
+        self.disk_saved = 0
+        self.mxids_to_exclude = mxids_to_exclude
+        self.excluded_by_user = {
+            mxids_to_exclude[i]: 0 for i in range(len(mxids_to_exclude))
+        }
+
+    def check_date(self, folder_name: str, applications_to_delete: List[str]) -> None:
```
richvdh, line 39: as with the properties: suggest giving the internal methods `_` prefixes.

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+        today = datetime.today()
+        for days_ago in self.days_to_check:
+            target = today - timedelta(days=days_ago)
+            folder_name = target.strftime("%Y-%m-%d")
+            applications = []
+            for name in self.limits.keys():
+                if self.limits[name] < days_ago:
+                    applications.append(name)
+            self.check_date(self.root_path + "/" + folder_name, applications)
+        pass
```
richvdh, line 118:
```suggestion
```

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+                print(f"I deleted {file}")
+                os.unlink(file)
+
+        if self.dry_run:
+            print(f"I would remove directory {rageshake_folder_path}")
+        else:
+            print(f"I removing directory {rageshake_folder_path}")
+            os.rmdir(rageshake_folder_path)
+
+    def cleanup(self) -> None:
```
richvdh, line 108: I would put the public entry point above the helper methods, so that I can read from top to bottom.

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+        today = datetime.today()
+        for days_ago in self.days_to_check:
+            target = today - timedelta(days=days_ago)
+            folder_name = target.strftime("%Y-%m-%d")
+            applications = []
+            for name in self.limits.keys():
```
richvdh, line 114: it looks like, if no `--limits` are provided, nothing is deleted? Is that deliberate?

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+            mxids_to_exclude[i]: 0 for i in range(len(mxids_to_exclude))
+        }
+
+    def check_date(self, folder_name: str, applications_to_delete: List[str]) -> None:
+        if len(applications_to_delete) == 0:
+            print(f"W Not checking {folder_name}, no applications would be removed")
+            return
+
+        # list folder_name for rageshakes:
+        # foreach:
```
richvdh, line 45:
```suggestion
```

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+        files = glob.iglob(folder_name + "/[0-9]*")
+
+        checked = 0
+        deleted = 0
+        for rageshake_name in files:
+            checked = checked + 1
+            if self.check_rageshake(rageshake_name, applications_to_delete):
+                deleted = deleted + 1
```
richvdh, line 53:
```suggestion
                deleted += 1
```

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+
+        print(
+            f"I Checked {folder_name} for {applications_to_delete},  deleted {deleted}/{checked} rageshakes"
```
richvdh, line 56: I think this would scare me in dryrun mode.

```suggestion
            f"I Checked {folder_name} for {applications_to_delete}, {'would delete' if self.dryrun else 'deleted'} {deleted}/{checked} rageshakes"
```

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+        )
+
+        self.deleted = self.deleted + deleted
+        self.checked = self.checked + checked
```
richvdh, line 60:
```suggestion
        self.deleted += deleted
        self.checked += checked
```

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+            app_name = None
+            mxid = None
+            with gzip.open(rageshake_folder_path + "/details.log.gz") as details:
+                for line in details.readlines():
+                    parts = line.decode("utf-8").split(":", maxsplit=1)
+                    if parts[0] == "Application":
+                        app_name = parts[1].strip()
+                    if parts[0] == "user_id":
+                        mxid = parts[1].strip()
+            print(f"app_name {app_name} user_id {mxid}")
```
richvdh, line 76: this doesn't seem to match the format of the other log lines. Maybe you meant to remove it?

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def cleanup(self) -> None:
+        today = datetime.today()
+        for days_ago in self.days_to_check:
+            target = today - timedelta(days=days_ago)
+            folder_name = target.strftime("%Y-%m-%d")
+            applications = []
+            for name in self.limits.keys():
+                if self.limits[name] < days_ago:
+                    applications.append(name)
```
richvdh, line 116: Given we're going to check `foo in applications` a lot, it might be good to make it a set:

```suggestion
            applications = set()
            for name in self.limits.keys():
                if self.limits[name] < days_ago:
                    applications.add(name)
```

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+            f"I Checked {folder_name} for {applications_to_delete},  deleted {deleted}/{checked} rageshakes"
+        )
+
+        self.deleted = self.deleted + deleted
+        self.checked = self.checked + checked
+        # optionally delete folder if we deleted 100% of rageshakes, but for now it' s fine.
+
+    def check_rageshake(
+        self, rageshake_folder_path: str, applications_to_delete: List[str]
+    ) -> bool:
```
richvdh, line 65: docstring: what does this return?

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+                for line in details.readlines():
+                    parts = line.decode("utf-8").split(":", maxsplit=1)
+                    if parts[0] == "Application":
+                        app_name = parts[1].strip()
+                    if parts[0] == "user_id":
+                        mxid = parts[1].strip()
+            print(f"app_name {app_name} user_id {mxid}")
+            if app_name in applications_to_delete:
+                if mxid in self.mxids_to_exclude:
+                    self.excluded_by_user[mxid] = self.excluded_by_user[mxid] + 1
```
richvdh, line 79:
```suggestion
                    self.excluded_by_user[mxid] += 1
```

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+        self.dry_run = dry_run
+        self.days_to_check = days_to_check
+        self.limits = limits
+        self.root_path = root_path
+        self.deleted = 0
+        self.checked = 0
+        self.disk_saved = 0
+        self.mxids_to_exclude = mxids_to_exclude
+        self.excluded_by_user = {
+            mxids_to_exclude[i]: 0 for i in range(len(mxids_to_exclude))
```
richvdh, line 36:
```suggestion
            mxid: 0 for mxid in mxids_to_exclude
```

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+            with gzip.open(rageshake_folder_path + "/details.log.gz") as details:
+                for line in details.readlines():
+                    parts = line.decode("utf-8").split(":", maxsplit=1)
+                    if parts[0] == "Application":
+                        app_name = parts[1].strip()
+                    if parts[0] == "user_id":
+                        mxid = parts[1].strip()
+            print(f"app_name {app_name} user_id {mxid}")
+            if app_name in applications_to_delete:
+                if mxid in self.mxids_to_exclude:
```
richvdh, line 78: If you do this, `mxids_to_exclude` is redundant:

```suggestion
                if mxid in self.excluded_by_user:
```

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+                    self.excluded_by_user[mxid] = self.excluded_by_user[mxid] + 1
+                else:
+                    self.delete(rageshake_folder_path)
+                    return True
+            return False
```
richvdh, line 83:
```suggestion
```

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+        return False
+
+    def delete(self, rageshake_folder_path: str) -> None:
+        files = glob.glob(rageshake_folder_path + "/*")
+        for file in files:
+            self.disk_saved += os.stat(file).st_size
+            if self.dry_run:
+                print(f"I would delete {file}")
+            else:
+                print(f"I deleted {file}")
```
richvdh, line 99:
```suggestion
                print(f"I deleting {file}")
```

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+        except FileNotFoundError as e:
+            print(
+                f"W Unable to open {e.filename} to check for application name. Ignoring this folder."
+            )
+
+        return False
+
+    def delete(self, rageshake_folder_path: str) -> None:
+        files = glob.glob(rageshake_folder_path + "/*")
+        for file in files:
```
richvdh, line 94: I'd probably use `scandir` for this, because (a) glob is a bit of a sledgehammer to return all the files (b) it avoids a separate `stat` on each file.

richvdh (changes requested):
looks good but oops I seem to have made a lot of comments!

scripts/cleanup.py
michaelkaye: I did run `mypy`, `black` and `isort` before i committed and there were no errors, i may have just driven mypy wrong. Will try to get it to error before i fix it.

michaelkaye: can definitely make mypy complain about other things in the file when I run it on commandline; however, pycharm complains correctly.. No idea; fixed anyway

scripts/cleanup.py
michaelkaye: Yes; i decided against a default delete or similar; It's not as bad as it sounds as we're just about to start limiting valid incoming application names so there will always be a limited number of items in the list

scripts/cleanup.py
michaelkaye: I understand how it makes the variable redundant; but I think I prefer to have the configuration of what to do `_mxids_to_exclude` separated from the side effect of reporting on what we excluded, `excluded_count_by_user`

I don't expect the list to end up being very long; maybe 10-20 mxids.

scripts/cleanup.py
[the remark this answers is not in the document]
michaelkaye: I'm not sure about this;

I was aiming for not taking any action (deletion etc) if the details.log.gz file was not found. Putting the remaining logic outside the `try/except` would mean it does get run, even if the (current) if block would always evaluate to False as app_name would always be None.

I feel it's more obvious if we simply don't evaluate the decision at all when we get a file not found error.

scripts/cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+        dest="path",
+        type=str,
+        required=True,
+        help="Root path of rageshakes (eg /home/rageshakes/bugs/)",
+    )
+
+    args = parser.parse_args()
+    application_limits: Dict[str, int] = {}
+    for x in args.limits:
+        application_limits[x.split(":")[0]] = int(x.split(":")[1])
```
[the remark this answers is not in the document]
michaelkaye: I don't think anything stops it other than humans not letting an app name contain `:` in allowed app names for the rageshake server.

scripts/cleanup.py
```diff
@@ -180,8 +202,8 @@ def main():
 
     args = parser.parse_args()
     application_limits: Dict[str, int] = {}
-    for x in args.limits:
-        parts = x.rsplit(":", 1)
+    for l in args.limits:
+        parts = l.rsplit(":", 1)
```
richvdh, line 206: `l` may be a poor name for a var due to the confusion with `I` and `1`, but :man_shrugging:

richvdh (approved):
lgtm!

PR #64, opened 2022-10-28
Migrate from Buildkite to GitHub actions

michaelkaye:
Re-implements the buildkite tests in GHA. LIke for like at the moment; have removed the specific scripts for buildkite as GHA allows in-repo editing of them. Have left the linting script alone as it's also used for local linting purposes.

Note: Buildkite test failures are expected as part of this PR tidies up the old buildkite scripts.


.github/workflows/linting.yaml
```diff
@@ -0,0 +1,48 @@
...
+    branches: [master]
+
+jobs:
+  changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          # checkout all branches; we only want origin/master, required to allow towncrier to diff
```
richvdh, line 15: > checkout all branches

> we only want origin/master

aren't these mutually contradictory?

richvdh (approved):
sure

PR #67, opened 2022-12-13
Block applications not listed in an allow list.

michaelkaye:
Optionally admins can now set a list of application names that will be accepted by this server. 

This is in reaction to admins spending time triaging bug reports for applications they don't own / can't do anything with.

The behaviour will stay as before (all applications permitted)  if the option is unset or an empty list; but will log a new warning on startup to say this behaviour is occuring.

If the report is blocked, the error message will include a link to the docs/blocked_rageshake.md file which contains some pointers on why and what to do afterwards; This isn't strictly neccessary but should ease peoples' debugging flows from that point.

### DIFF
--- a/docs/blocked_rageshake.md
+++ b/docs/blocked_rageshake.md
@@ -27,7 +27,7 @@ There is generally a configuration file in your application that you can alter t
 
 The easiest solution to this error is to stop sending rageshakes entirely, which may require a code or configuration change in your client.
 
-However, if you wish to accept bug reports from your users applications; you will need to run your own copy of this rageshake server and update the URL appropriately.
+However, if you wish to accept bug reports from your users applications, you will need to run your own copy of this rageshake server and update the URL appropriately.
 
 ## Application specific config locations:
  * element-web: `bug_report_endpoint_url` in the [sample configuration for element web](https://github.com/vector-im/element-web/blob/develop/config.sample.json).
--- a/docs/blocked_rageshake.md
+++ b/docs/blocked_rageshake.md
@@ -1,0 +1,35 @@
+# Rageshake server not accepting rageshakes
+
+This page contains information useful to someone who has had their rageshake rejected by a rageshake server.
+
+We include it within the error messages to provide a place with context for users reading the error message and wanting
+to know more.
+
+## For matrix client users
+
+Thank you for attempting to report a bug with your matrix client; unfortunately your client application is likely incorrectly configured.
+
+The rageshake server you attempted to upload a report to is not accepting rageshakes from your client at this time.
+
+Generally, the developers who run a rageshake server will only be able to handle reports for applications they are developing,
+and your application is not listed as one of those applications.
+
+Please contact the distributor of your application or the administrator of the web site you visit to report this as a problem.
+
+## For developers of matrix clients
+
+Your application is likely based on one of the matrix SDKs or element applications, if it is submitting rageshakes to a rageshake server.
+
+A change has been made to pre-filter reports that the developers using this rageshake server for applications they do not have control over.
+Typically reports from unknown applications would have to be manually triaged and discarded; there is now automatic filtering in place, which reduces overall effort.
+
+There is generally a configuration file in your application that you can alter to change where these reports are sent, which may require rebuilding and releasing the client.
+
+The easiest solution to this error is to stop sending rageshakes entirely, which may require a code or configuration change in your client.
+
+However, if you wish to accept bug reports from your users applications; you will need to run your own copy of this rageshake server and update the URL appropriately.
+
+## Application specific config locations:
+ * element-web: `bug_report_endpoint_url` in the [sample configuration for element web](https://github.com/vector-im/element-web/blob/develop/config.sample.json).
+ * element-ios: `bugReportEndpointUrlString` in the [BuildSettings.swift](https://github.com/vector-im/element-ios/blob/develop/Config/BuildSettings.swift)
+ * element-android: `bug_report_url` in the [config.xml file for the build](https://github.com/vector-im/element-android/blob/develop/vector-config/src/main/res/values/config.xml)
--- a/main.go
+++ b/main.go
@@ -48,6 +48,10 @@ type config struct {
 	// External URI to /api
 	APIPrefix string `yaml:"api_prefix"`
 
+	// Allowed rageshake app names
+	AllowedAppNames   []string `yaml:"allowed_app_names"`
+	AllowedAppNameMap map[string]bool
+
 	// A GitHub personal access token, to create a GitHub issue for each report.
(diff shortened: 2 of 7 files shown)


main.go
```diff
@@ -48,6 +48,10 @@ type config struct {
 	// External URI to /api
 	APIPrefix string `yaml:"api_prefix"`
 
+	// Allowed rageshake app names
+	AllowedAppNames   []string `yaml:"allowed_app_names"`
+	AllowedAppNameMap map[string]bool
```
t3chguy, line 53: this probably doesn't need exporting given it is just a memoized helper to quicker lookups

docs/blocked_rageshake.md
```diff
@@ -0,0 +1,35 @@
...
+Your application is likely based on one of the matrix SDKs or element applications, if it is submitting rageshakes to a rageshake server.
+
+A change has been made to pre-filter reports that the developers using this rageshake server for applications they do not have control over.
+Typically reports from unknown applications would have to be manually triaged and discarded; there is now automatic filtering in place, which reduces overall effort.
+
+There is generally a configuration file in your application that you can alter to change where these reports are sent, which may require rebuilding and releasing the client.
+
+The easiest solution to this error is to stop sending rageshakes entirely, which may require a code or configuration change in your client.
+
+However, if you wish to accept bug reports from your users applications; you will need to run your own copy of this rageshake server and update the URL appropriately.
```
richvdh, line 30:
```suggestion
However, if you wish to accept bug reports from your users applications, you will need to run your own copy of this rageshake server and update the URL appropriately.
```

main.go
```diff
@@ -48,6 +48,10 @@ type config struct {
 	// External URI to /api
 	APIPrefix string `yaml:"api_prefix"`
 
+	// Allowed rageshake app names
+	AllowedAppNames   []string `yaml:"allowed_app_names"`
+	allowedAppNameMap map[string]bool
```
richvdh, line 53: I'm a bit unconvinced that we should be putting derived properties in this structure; in my opinion it represents exactly the settings we read from the file, and if we want to preprocess settings, then that should go elsewhere.

My approach to this would be to add `allowedAppNameMap` as a separate property in [`submitServer`](https://github.com/matrix-org/rageshake/blob/e61c48177d4c40114abae9dcc31b277494979eaf/submit.go#L49). To populate it, I would create a new function `NewSubmitServer` which makes a new `submitServer`.

submit.go
```diff
@@ -273,35 +285,13 @@ func parseJSONRequest(w http.ResponseWriter, req *http.Request, reportDir string
 		}
 	}
 
-	// backwards-compatibility hack: current versions of riot-android
-	// don't set 'app', so we don't correctly file github issues.
```
richvdh, line 277: +1 to ripping this out, though ideally it would have been a separate PR. Never mind now.

main.go
michaelkaye: Ah yes, having looked at it again, it looks like the rest of the configuration that is generated is done when passing it into the creation of a submitServer struct, it seems OK to just add another option [in there](https://github.com/matrix-org/rageshake/pull/67/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R161) - i don't think we need to make a function for just this part of the problem; other than to factor it out of the main method (which is actually one big NewSubmitServer function).

main.go
```diff
@@ -184,6 +186,16 @@ func main() {
...
+	if len(cfg.AllowedAppNames) == 0 {
+		fmt.Println("Warning: allowed_app_names is empty. Accepting requests from all app names")
+	}
+	var allowedAppNameMap = make(map[string]bool)
+	for _, app := range cfg.AllowedAppNames {
+		allowedAppNameMap[app] = true
+	}
+	return allowedAppNameMap
+}
 func configureGenericWebhookClient(cfg *config) *http.Client {
```
richvdh, line 199:
```suggestion

func configureGenericWebhookClient(cfg *config) *http.Client {
```

submit.go
```diff
@@ -192,7 +193,7 @@ func (s *submitServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	}
 
 	// Filter out unwanted rageshakes, if a list is defined
-	if len(s.cfg.AllowedAppNames) != 0 && !s.cfg.allowedAppNameMap[p.AppName] {
+	if !s.allowedAppNameMap[p.AppName] {
```
richvdh, line 196: this means we will not accept any rageshakes if the list is unconfigured (in contrast to what the printed warning and documentation say)?

main.go
```diff
@@ -201,4 +217,5 @@ func loadConfig(configPath string) (*config, error) {
 		return nil, err
 	}
 	return &cfg, nil
+
```
richvdh, line 220:
```suggestion
```

main.go
```diff
@@ -149,10 +152,13 @@ func main() {
...
 		apiPrefix = strings.TrimRight(apiPrefix, "/")
 	}
+
+	appNameMap := configureAppNameMap(cfg)
+
 	log.Printf("Using %s/listing as public URI", apiPrefix)
 
 	rand.Seed(time.Now().UnixNano())
-	http.Handle("/api/submit", &submitServer{ghClient, glClient, apiPrefix, slack, genericWebhookClient, cfg})
+	http.Handle("/api/submit", &submitServer{ghClient, glClient, apiPrefix, slack, genericWebhookClient, appNameMap, cfg})
```
richvdh, line 161: > - i don't think we need to make a function for just this part of the problem; other than to factor it out of the main method (which is actually one big NewSubmitServer function).

I mean yes, there's a whole lot of stuff that could do with factoring out to a function that just creates the SubmitServer without calling `http.Handle` etc. But it's fine to leave this for now.

richvdh (changes requested):

submit.go
michaelkaye: You're right - fixed, and patch now contains a (simple) test that triggers the three cases by creating a submitServer with no config.

michaelkaye:
Apologies for all the to and fro with this; your feedback is appreciated :)

submit_test.go
```diff
@@ -56,6 +56,56 @@ func testParsePayload(t *testing.T, body, contentType string, tempDir string) (*
...
+	}
+	if submitSimpleRequestToServer(t, invalidAppNameMap, body) != 400 {
+		t.Fatal("nonmatching app was not rejected")
+	}
+
+	emptyAppNameMap := make(map[string]bool)
+	if submitSimpleRequestToServer(t, emptyAppNameMap, body) != 200 {
+		t.Fatal("empty map did not allow all")
+	}
+
```
richvdh, line 106:
```suggestion
```

richvdh (approved):
lgtm!

PR #70, opened 2023-03-21
Extra automation for release process (building binaries and docker containers)

### LINKED ISSUE #51: Latest release did not tag the right docker images
Latest release tag was v1.3, but the Docker CI expects a semver-compatible tag, e.g. vX.Y.Z

We should:
 - do a proper v1.3.0 release/tag
 - update the [releasing docs](https://github.com/matrix-org/rageshake/blob/29c2f9a48fbe04299a9f20e8790114ae8e7c1f6e/RELEASING.md) to reflect that tag format

### LINKED ISSUE #69: No container tags corresponding to version releases
It looks like there's no `v1.8` (or `1.8`) container image tag published to the `ghcr.io/matrix-org/rageshake` container registry.

It's just `master`, etc., which is a moving target and dangerous to use.

michaelkaye:
Updates the release process such that creation of a github Release will be enough to trigger the docker image build and the rageshake.tar.gz to be uploaded to the release.

Requires that we move to a full "semver" semantic version in order to trigger the various tools to work; so no more `v1.8` ; we should use  `v1.9.0`.

Fixes #69  
Fixes #51

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -60,7 +60,7 @@ jobs:
         run: tar -czf rageshake.tar.gz rageshake
       - name: Upload tarball to matching release
         if: github.event.release
-        uses: softprops/action-gh-release@v1
+        uses: softprops/action-gh-release@de2c0eb89ae2a093876385947365aca7b0e5f844
         with:
           files: rageshake.tar.gz
         env:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,9 +1,18 @@
-name: Linting and checks
+name: Linting, build, test and release
+
+# Runs on each PR to lint and test
+
+# Runs on each release (via github UI) to lint and test, then upload binary to the release
 
 on:
   pull_request:
   push:
     branches: [master]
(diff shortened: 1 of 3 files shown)


richvdh (approved):
lgtm otherwise

richvdh:
> Not sure if we have any worries about doing this build in GHA vs locally on a machine; it seems less fragile to do so (eg, ensuring we do not build from different sources than we expected) ; but does make the build out of the direct control of a developer.

I think our general policy on this is: building in a reproducible, clean environment (GHA) is a win over building on dev machines. There is a risk of either GHA or a dev machine being compromised and hence building a compromised binary, but I have more confidence in that being noticed on GHA than a dev machine.

michaelkaye:
Am going to do a release of 1.9.0 with minimal changes (it does include that new script) to test the release flow and generate at least one tagged docker image for #69

michaelkaye:
This isn't perfect; in docker, we've tagged the 1.9.0-debug version as 'latest' when we probably want to tag 1.9.0 as latest. Will work out how to manage that away.

PR #72, opened 2023-03-24
Enhance scripts/cleanup.py to be able to run as a continual task.

michaelkaye:
This is useful in the situation where you want to run this as a sidecar in a k8s pod, because while k8s has the concept of a CronJob - it doesn't run within the same pod; so you'd end up with two pods both needing to mount a block device (say, where you keep the bugs) read-write; which I don't believe works well. 

Have tested it with an invocation like this to test every 0.01 hours, which ran as expected
`michaelk@michaelk-XPS-13-9370:~/work/rageshake$ ./scripts/cleanup.py --repeat-delay-hours 0.01 --max-days 100 --path /home/michaelk/bugs/ --dry-run element-web:10` 

The following both run exactly once, which seems valid.

`michaelk@michaelk-XPS-13-9370:~/work/rageshake$ ./scripts/cleanup.py --repeat-delay-hours 0 --max-days 100 --path /home/michaelk/bugs/ --dry-run element-web:10` 
and
`michaelk@michaelk-XPS-13-9370:~/work/rageshake$ ./scripts/cleanup.py --max-days 100 --path /home/michaelk/bugs/ --dry-run element-web:10` 

While I haven't done a comprehensive check for leaks, running this over a period of time shows memory usage as flat, and lsof didn't reveal any file descriptors being held onto incorrectly, so I think the continual recreation of the Cleanup object is keeping things tidy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,11 +120,15 @@ Details on the request and expected response are [available](docs/generic\_webho
 
 ## Cleanup script
 
-A python script is provided in scripts/cleanup.py and in a docker container. It can be configured
-using the commandline options to filter rageshakes in a given date range.
+A python script is provided in scripts/cleanup.py and in a
+[docker container](https://github.com/orgs/matrix-org/packages/container/package/rageshake%2Fscripts).
+It can be configured using the commandline options available via `cleaup.py --help`.
 
 It can either be run via a cronjob at appropriate intervals (typically daily), or
 be set to run in a continual mode with something like `--repeat-delay-hours 24`
 to repeat running after approximately 24 hours.
 
-Note that this script will scan all logs older than the smallest configured retention period, up to the limit specified by --max-days. This can be an IO and CPU intensive process. 
+Note that this script will scan all logs older than the smallest configured retention period,
+up to the limit specified by `--max-days` or each of the days in `--days-to-check`.
+This can be an IO and CPU intensive process if a large number of files are scanned.
+
--- a/README.md
+++ b/README.md
@@ -127,5 +127,4 @@ It can either be run via a cronjob at appropriate intervals (typically daily), o
 be set to run in a continual mode with something like `--repeat-delay-hours 24`
 to repeat running after approximately 24 hours.
 
-Be aware that each run will scan all files between --max-days and the smallest retention period set,
-ensuring we always have up to date information, however this may be an intensive process.
+Note that this script will scan all logs older than the smallest configured retention period, up to the limit specified by --max-days. This can be an IO and CPU intensive process. 
--- a/README.md
+++ b/README.md
@@ -121,11 +121,10 @@ Details on the request and expected response are [available](docs/generic\_webho
 ## Cleanup script
 
 A python script is provided in scripts/cleanup.py and in a docker container. It can be configured
-using the commandline options to filter rageshakes in a given date range. It is expected to be
-run as a cronjob on a daily basis.
+using the commandline options to filter rageshakes in a given date range.
 
-The script is idempotent enough to be re-runnable on a more frequent basis if required. If a cronjob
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -117,3 +117,16 @@ issues created.
 
 Details on the request and expected response are [available](docs/generic\_webhook.md).
 
+
+## Cleanup script
+
+A python script is provided in scripts/cleanup.py and in a docker container. It can be configured
```
richvdh, line 123: > in a docker container

any clues as to where I can find this mythical docker container?

scripts/cleanup.py
```diff
@@ -200,7 +201,33 @@ def main():
...
+    parser.add_argument(
+        "--repeat-delay-hours",
+        dest="repeat_delay_hours",
+        type=float,
+        help="After each execution, wait this number of hours then run again. An alternative to configuring a cronjob for ongoing cleanup.",
+    )
+
     args = parser.parse_args()
+
+    if args.repeat_delay_hours:
```
richvdh, line 213: to distinguish from `--repeat-delay-hours=0`:

```suggestion
    if args.repeat_delay_hours is defined:
```

scripts/cleanup.py
```diff
@@ -200,7 +201,33 @@ def main():
         help="Root path of rageshakes (eg /home/rageshakes/bugs/)",
     )
 
+    parser.add_argument(
+        "--repeat-delay-hours",
+        dest="repeat_delay_hours",
+        type=float,
+        help="After each execution, wait this number of hours then run again. An alternative to configuring a cronjob for ongoing cleanup.",
```
richvdh, line 208: Possibly clearer, ymmv

```suggestion
        help="""
            Instead of exiting after execution, run repeatedly, waiting this number of hours between each execution.
            An alternative to configuring a cronjob for ongoing cleanup.
            """,
```

scripts/cleanup.py
```diff
@@ -200,7 +201,33 @@ def main():
...
+            time.sleep(args.repeat_delay_hours * 60 * 60)
+    else:
+        execute(args)
+
+
+def execute(args) -> None:
+    """
+    Creation, configuration and execution of a cleanup task based on args.
+
+    Allows exceptions to propigate to the caller for handling.
```
richvdh, line 228:
```suggestion
    Allows exceptions to propagate to the caller for handling.
```

README.md
```diff
@@ -117,3 +117,16 @@ issues created.
...
+using the commandline options to filter rageshakes in a given date range. It is expected to be
+run as a cronjob on a daily basis.
+
+The script is idempotent enough to be re-runnable on a more frequent basis if required. If a cronjob
+cannot be configured, it can be run in a continual mode with something like `--repeat-delay-hours 24`
+to repeat running after approximately 24 hours.
```
richvdh, line 129: Is there any particular reason to prefer a cronjob?

```suggestion
using the commandline options to filter rageshakes in a given date range.

It can either be run via a cronjob at appropriate intervals (typically daily), or
be set to run in a continual mode with something like `--repeat-delay-hours 24`
to repeat running after approximately 24 hours.
```

README.md
```diff
@@ -117,3 +117,16 @@ issues created.
...
+
+Be aware that each run will scan all files between --max-days and the smallest retention period set,
+ensuring we always have up to date information, however this may be an intensive process.
```
richvdh, line 132: I'm not quite sure what this is trying to tell me. I think it's something like this:

```suggestion
Note that this script will scan all logs older than the smallest configured retention period (up to `--max-days`).
This can be an IO-intensive process.
```

richvdh (commented):
basically looks good, but a few nits

michaelkaye:
Have reworded a few of the review comments (ones left unresolved) but accepted most as-is.

richvdh (approved):
lgtm

PR #76, opened 2023-04-21
Third Room submit report docs

ajbura:


docs/submitted_reports.md
```diff
@@ -85,3 +85,36 @@ log-[1-7].log
 ```
 
 Log 1-7 are logs from a round-robin buffer and are ordered but the start point is undefined
+
+## Third Room
+
+Third Room uses `logs.json` file to submit logs.
```
richvdh, line 91:
```suggestion
Third Room submits logs as a single `logs.json` file.
```

Is that right?

docs/submitted_reports.md
```diff
@@ -85,3 +85,36 @@ log-[1-7].log
...
+  l: number; // log level
+  e?: { // error
+      stack?: string;
+      name: string;
+      message: string;
+  };
+  f?: boolean; // forced
+  c?: LogItem[]; // children
+}
+```
```
richvdh, line 120: I don't find this particularly easy to understand (also it assumes knowledge of typescript syntax, which not all readers of this file may know).

Could you maybe write it out as a markdown list, and expand on the meaning of the properties

```markdown
Each `LogItem` is an object with the following properties:

 * `s`: start time XXX what units?
 * `d`: duration XXX of what? When is this present? When is it not?
```

etc

richvdh (changes requested):

richvdh (approved):
much clearer, thanks!

PR #77, opened 2023-05-02
Update release GHA to run on publish of release, not creation.

michaelkaye:


michaelkaye:
Pushing this through because we've got a broken release with no tarball atm - if i merge and unpublish/publish the release again, it should fix things.

michaelkaye:
(it did)

PR #81, opened 2023-08-01
add link to archive in the submitted issue

bnjbvr:
This adds a link to the full archive of the rageshake logs, avoiding a few manual steps to download them. For what it's worth, the server will happily respond with the full archive even in the presence of a trailing slash in the `listingURL` (it might be an HTTP thing).

### DIFF
--- a/changelog.d/81.feature
+++ b/changelog.d/81.feature
@@ -1,0 +1,1 @@
+Add a link to the archive containing all the logs in the webhook's body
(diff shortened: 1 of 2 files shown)


bnjbvr:
@richvdh Gentle review ping :)

changelog.d/81.feature
```diff
@@ -0,0 +1 @@
+Add a link to the archive containing all the logs in the webhook's body
```
richvdh, line 1:
```suggestion
Add a link to the archive containing all the logs in the issue body.
```

richvdh (approved):
sorry, I was away.

LGTM otherwise, thanks!

By the way, note we don't use conventional commits in this project.

PR #84, opened 2024-03-05
Generate emails and github/gitlab issue bodies from a template

richvdh:
The driver for this is that I want to add a link to our internal rageshake-search tool to the body of reported issues. Rather than add more magic special-casing that, making the format of issue bodies more configurable seems a good thing.

### DIFF
--- a/main.go
+++ b/main.go
@@ -28,6 +28,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"text/template"
 	"time"
 
 	"github.com/google/go-github/github"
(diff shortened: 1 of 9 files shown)


submit.go
```diff
@@ -665,46 +687,47 @@ func buildReportBody(p payload, newline, quoteChar string) *bytes.Buffer {
 	return &bodyBuf
 }
 
-func buildGenericIssueRequest(p payload, listingURL string) (title, body string) {
-	bodyBuf := buildReportBody(p, "  \n", "`")
```
kegsay, line 669: `buildReportBody` should be removed, but I see it's still being used for sending email notifications. Was there any reason not to use the template for emails as well?

rageshake.sample.yaml
```diff
@@ -21,6 +21,19 @@ github_token: secrettoken
 github_project_mappings:
    my-app: octocat/HelloWorld
 
+# A template for the body of Github and Gitlab issues. The default template is as shown below.
+#
+# See `README.md` for more information on what can be specified here.
+issue_body_template: |
+  {{ .UserText }}
```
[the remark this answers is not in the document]
richvdh: Not stepping on the YAML parser is pretty easy; you just have to indent each line. You're not wrong, though.

submit.go
richvdh: Well, it opens up a can of worms. Should we use the same template for both notifications? If so, the template has to handle the quote chars. If not, you've got two templates to keep in sync.

That said, keeping two templates in sync is arguably easier than keeping a template and a hard-coded impl in sync, so maybe that answers my question...

main.go
```diff
@@ -212,16 +211,26 @@ func main() {
...
-	issueTemplate := cfg.IssueBodyTemplate
-	if issueTemplate == "" {
-		issueTemplate = DefaultIssueBodyTemplate
+// parseTemplate parses a template file, with fallback to default.
+//
+// If `configFileSettingValue` is non-empty, it is used as the name of a file to read. Otherwise, `defaultTemplate` is
+// used.
+//
+// The template text is then parsed into a template named `templateName`.
+func parseTemplate(defaultTemplate string, configFileSettingValue string, templateName string) *template.Template {
```
kegsay, line 220: `configFileSettingValue` feels like a bit of a mouthful. Do we really care that the template file path comes from the config? I'd rename this to `templateFilePath` personally.

kegsay (approved):
LGTM!

PR #85, opened 2024-05-09
Add support for blocking specific app/version/label combinations

kegsay:


submit.go
```diff
@@ -99,7 +99,7 @@ type issueBodyTemplatePayload struct {
 type genericWebhookPayload struct {
 	payload
 	// If a github/gitlab report is generated, this is set.
-	ReportURL  string `json:"report_url"`
+	ReportURL string `json:"report_url"`
```
richvdh, line 102: non-blocking comment: any idea why this is happening? Has `gofmt` changed the rules about struct formatting?I'm not convinced that the updated style is an improvement, but I don't care that much as long as we're not just going to flip-flop between the two.

submit.go
```diff
@@ -422,11 +434,11 @@ func formPartToPayload(field, data string, p *payload) {
 
 // we use a quite restrictive regexp for the filenames; in particular:
 //
-// * a limited set of extensions. We are careful to limit the content-types
-//   we will serve the files with, but somebody might accidentally point an
-//   Apache or nginx at the upload directory, which would serve js files as
-//   application/javascript and open XSS vulnerabilities. We also allow gzipped
-//   text and json on the same basis (there's really no sense allowing gzipped images).
+//   - a limited set of extensions. We are careful to limit the content-types
```
richvdh, line 437: :-1:. This was a bullet list of items starting with `*`. If you want to change it to start with `-` that's ok, but please do all the items, not just the first.

main.go
```diff
@@ -63,6 +64,8 @@ type config struct {
 	// Allowed rageshake app names
 	AllowedAppNames []string `yaml:"allowed_app_names"`
 
+	RejectionConditions []RejectionCondition `yaml:"rejection_conditions"`
```
richvdh, line 67: comment please

main.go
```diff
@@ -93,6 +96,54 @@ type config struct {
 	GenericWebhookURLs []string `yaml:"generic_webhook_urls"`
 }
 
+// RejectionCondition contains the fields that should match a bug report for it to be rejected.
+type RejectionCondition struct {
+	Version string `yaml:"version"`
+	Label   string `yaml:"label"`
+	App     string `yaml:"app"`
```
richvdh, line 103: I note that the behaviour of `App==""` is quite different to the behaviour if `Version==""` or `Label=""`. Please document those behaviours here.

(should we instead use `*string` for Label and App, so that `nil` is different to `""` ?)

rageshake.sample.yaml
```diff
@@ -12,6 +12,16 @@ listings_auth_pass: secret
 # An empty or missing list will retain legacy behaviour and permit reports from any application name.
 allowed_app_names: []
 
+# If any submission matches one of these rejection conditions, the submission is rejected.
```
richvdh, line 15: It would be good to mention here that `app` is required, but `version` and `label` are both optional.

richvdh (changes requested):

submit.go
kegsay: Unsure why this is happening tbh. Empirically, this is what `go fmt .` does, so it's not an IDE thing. 
```
go version
go version go1.22.1 darwin/arm64
```

submit.go
kegsay: This was `go fmt` not me. Bullet points are `-` and shouldn't have new line separations which is probably why it didn't see the ones further down.

submit.go
richvdh: same for me. weird.

main.go
kegsay: There is no distinction between nil and "" in the filter because it doesn't make any practical difference as 0-length versions/labels are nonsensical. Adding `*string` just forces us to do more nil checks.

submit.go
kegsay: https://tip.golang.org/doc/go1.19#go-doc are you on an old version?

richvdh: No, I'm saying that I'm on go1.22, and it does the same as you. It's entirely possible the current formatting came from an earlier version of go. Thanks for hunting down the changes.

rageshake.sample.yaml
```diff
@@ -12,6 +12,18 @@ listings_auth_pass: secret
...
+# The 'app' field is required, but 'version' and 'label' are both optional. A condition with just
+# an 'app' will reject those apps, effectively acting as a blocklist for app in contrast to allowed_app_names.
```
richvdh, line 17:
```suggestion
# an 'app' will reject that app altogether, effectively acting as a blocklist for app in contrast to allowed_app_names.
```

richvdh (approved):
lgtm

PR #86, opened 2024-11-22
Parse user agent into a human-friendly string on submission

t3chguy:
I've had to look up UA's far too many times

### DIFF
--- a/submit.go
+++ b/submit.go
@@ -23,6 +23,7 @@ import (
 	"encoding/base32"
 	"encoding/json"
 	"fmt"
+	"github.com/ua-parser/uap-go/uaparser"
 	"io"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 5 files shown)


submit.go
```diff
@@ -321,7 +329,8 @@ func parseJSONRequest(w http.ResponseWriter, req *http.Request, reportDir string
 	parsed.AppName = p.AppName
 
 	if p.UserAgent != "" {
-		parsed.Data["User-Agent"] = p.UserAgent
+		parsed.Data["User-Agent"] = parseUserAgent(p.UserAgent)
+		parsed.Data["Raw-User-Agent"] = p.UserAgent
```
richvdh, line 333: This feels slightly magical to me: I think I'd prefer to keep `User-Agent` as the raw string, and have (say) `Parsed-User-Agent` for the parsed version.

submit.go
```diff
@@ -291,6 +292,13 @@ func parseRequest(w http.ResponseWriter, req *http.Request, reportDir string) *p
 	return p
 }
 
+var uaParser *uaparser.Parser = uaparser.NewFromSaved()
```
richvdh, line 295: https://pkg.go.dev/github.com/ua-parser/uap-go/uaparser#NewFromSaved is sadly lacking. I guess it loads from a preset list of user-agents.

richvdh (commented):
LGTM in general. I have slight reservations about feeding untrusted data into a library which doesn't seem to have a documented security posture, but I guess it's fine.

richvdh (approved):
lgtm

PR #88, opened 2025-01-30
Reject user texts matching a regex

davidegirardi:
Sometimes users send something they really should not in their rageshake descriptions. I propose here an approach to reject rageshakes whose description matches a regex. The rejection reason is reflected client-side so the user's client can show them want went wrong.

The sample config file provides a regex to reject user texts that could contain a recovery key.

TODO if we like this approach:
- [x] Update `docs/blocked_rageshake.md`
- [x] Update `main_test.go`
- <del>Update `submit_test.go`</del>
- [x] Update changelog

### DIFF
--- a/main.go
+++ b/main.go
@@ -110,7 +110,7 @@ type RejectionCondition struct {
 	// Message sent by the user, checked only if not empty
 	UserTextMatch string `yaml:"usertext"`
 	// Send this text to the client-side to inform the user why the server rejects the rageshake. Uses a default generic value if empty.
-	UserTextMatchReason string `yaml:"reason"`
+	Reason string `yaml:"reason"`
 }
 
 // shouldReject returns true if all the App, Version, Label and UserTextMatch attributes of a RejectionCondition match
@@ -118,8 +118,8 @@ func (c RejectionCondition) shouldReject(appName, version string, labels []strin
 	// Reject by default and then accept as soon as one condition does not match
 	reject = true
 	defaultReason := "app or user text rejected"
-	if c.UserTextMatchReason != "" {
-		reason = &c.UserTextMatchReason
+	if c.Reason != "" {
+		reason = &c.Reason
 	} else {
 		reason = &defaultReason
 	}
--- a/main.go
+++ b/main.go
@@ -113,72 +113,53 @@ type RejectionCondition struct {
 	Reason string `yaml:"reason"`
 }
 
-func (c RejectionCondition) rejectApp(p *payload, reason *string) *string {
-	// Use a form of short-circuit: if the reason pointer is already nil, some other rejection condition already failed and we don't need to perform any check
-	if reason != nil {
-		if c.App != "" && c.App != p.AppName {
-			reason = nil
-		}
-	}
-	return reason
+func (c RejectionCondition) matchesApp(p *payload) bool {
+	// Empty `RejectionCondition.App` is a wildcard which matches anything
+	return c.App == "" || c.App == p.AppName
 }
 
-func (c RejectionCondition) rejectVersion(p *payload, reason *string) *string {
-	if reason != nil {
-		version := ""
-		if p.Data != nil {
-			version = p.Data["Version"]
-		}
-		if c.Version != "" && c.Version != version {
-			reason = nil
-		}
+func (c RejectionCondition) matchesVersion(p *payload) bool {
+	version := ""
+	if p.Data != nil {
+		version = p.Data["Version"]
 	}
-	return reason
+	// Empty `RejectionCondition.Version` is a wildcard which matches anything
+	return c.Version == "" || c.Version == version
 }
 
-func (c RejectionCondition) rejectLabel(p *payload, reason *string) *string {
-	if reason != nil {
-		if c.Label != "" {
-			labelMatch := false
-			for _, l := range p.Labels {
-				if l == c.Label {
-					labelMatch = true
-					break
-				}
-			}
-			if !labelMatch {
-				reason = nil
-			}
+func (c RejectionCondition) matchesLabel(p *payload) bool {
+	// Empty `RejectionCondition.Label` is a wildcard which matches anything
+	if c.Label == "" {
+		return true
+	}
+	// Otherwise return true only if there is a label that matches
+	labelMatch := false
+	for _, l := range p.Labels {
+		if l == c.Label {
+			labelMatch = true
+			break
 		}
 	}
-	return reason
+	return labelMatch
 }
 
-func (c RejectionCondition) rejectUserText(p *payload, reason *string) *string {
-	if reason != nil {
-		if c.UserTextMatch != "" {
-			var userTextRegexp = regexp.MustCompile(c.UserTextMatch)
-			if !userTextRegexp.MatchString(p.UserText) {
-				reason = nil
-			}
+func (c RejectionCondition) matchesUserText(p *payload) bool {
+	// Empty `RejectionCondition.UserTextMatch` is a wildcard which matches anything
+	return c.UserTextMatch == "" || regexp.MustCompile(c.UserTextMatch).MatchString(p.UserText)
+}
+
+func (c RejectionCondition) shouldReject(p *payload) *string {
+	if c.matchesApp(p) && c.matchesVersion(p) && c.matchesLabel(p) && c.matchesUserText(p) {
+		// RejectionCondition matches all of the conditions: we should reject this submission/
+		defaultReason := "app or user text rejected"
+		if c.Reason != "" {
+			return &c.Reason
+		} else {
+			return &defaultReason
 		}
-	}
-	return reason
-}
-
-// shouldReject returns a rejection reason if the payload matches the condition, nil if the server should accept the rageshake
-func (c RejectionCondition) shouldReject(p *payload) (reason *string) {
-	defaultReason := "app or user text rejected"
-	if c.Reason != "" {
-		reason = &c.Reason
 	} else {
-		reason = &defaultReason
+		return nil
 	}
-	reason = c.rejectApp(p, reason)
-	reason = c.rejectLabel(p, reason)
-	reason = c.rejectUserText(p, reason)
-	reason = c.rejectVersion(p, reason)
-	return
 }
 
 func (c *config) matchesRejectionCondition(p *payload) (reason *string) {
--- a/main.go
+++ b/main.go
@@ -99,13 +99,13 @@ type config struct {
 }
 
 // RejectionCondition contains the fields that can match a bug report for it to be rejected.
+// All the (optional) fields must match for the rejection condition to apply
 type RejectionCondition struct {
-	// All the optional fields must match for the rejection condition to apply
 	// App name, checked only if not empty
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -98,54 +99,76 @@ type config struct {
 }
 
 // RejectionCondition contains the fields that should match a bug report for it to be rejected.
+// The condition matches a specific app, with optionall version and label, or the user submitted text
```
richvdh, line 102:
```suggestion
// The condition matches a specific app, with optional version and label, or the user submitted text
```

main.go
```diff
@@ -98,54 +99,76 @@ type config struct {
...
 type RejectionCondition struct {
-	// Required field: if a payload does not match this app name, the condition does not match.
+	// If a payload is not a UserTextMatch and does not match this app name, the condition does not match.
```
richvdh, line 104: This feels very confusing to reason about.

Shouldn't we just say that, for the condition to match, it must match **all** of `app`, `version`, `label`, `usertest` that are given? (and obviously, `app` should be optional)

main.go
```diff
@@ -98,54 +99,76 @@ type config struct {
...
 	App string `yaml:"app"`
 	// Optional: version that must also match in addition to the app and label. If empty, does not check version.
 	Version string `yaml:"version"`
 	// Optional: label that must also match in addition to the app and version. If empty, does not check label.
 	Label string `yaml:"label"`
+	// Alternative to matching app names, match the content of the user text
+	UserTextMatch       string `yaml:"usertext"`
+	UserTextMatchReason string `yaml:"reason"`
```
richvdh, line 112: can we make `reason`  more general: ie, make it apply even if `UserTextMatch` is not given? and if `reason` is unspecified, we make up something sensible depending on what conditions are specified?

main.go
```diff
@@ -98,54 +99,76 @@ type config struct {
...
-	}
-
-	// label was a condition and no label matches it => accept it
-	if c.Label != "" {
-		labelMatch := false
-		for _, l := range labels {
-			if l == c.Label {
-				labelMatch = true
-				break
+func (c RejectionCondition) shouldReject(appName, version string, labels []string, userText string) (reject bool, reason string) {
```
richvdh, line 120: Maybe I've been writing Rust too long, but I don't love the fact that `(false, "blah")` is a meaningless response.

Could maybe just return `*string` here, with `nil` meaning "do not reject" and anything else being the reason to reject.

richvdh (changes requested):
looks sensible in principle. A few thoughts.

main.go
davidegirardi: I wanted to be back-compatible but indeed making all elements optional will not break backward compatibility and make reasoning much easier.

main.go
davidegirardi: Yes. I think the approach you propose over will take care of this.

main.go
davidegirardi: Fixed in afd59fd11d9b6c976ef19ed25753484ddcc38651.

main.go
davidegirardi: Done in 20d4c24eb0ea4dc35c9117e66bbe1a06f4459670.

davidegirardi:
I think it's ready for review again. I abused the Data attribute to add an expected rejection reason in the tests. I don't know if that's OK.

main.go
```diff
@@ -97,28 +98,40 @@ type config struct {
 	GenericWebhookURLs []string `yaml:"generic_webhook_urls"`
 }
 
-// RejectionCondition contains the fields that should match a bug report for it to be rejected.
+// RejectionCondition contains the fields that can match a bug report for it to be rejected.
 type RejectionCondition struct {
-	// Required field: if a payload does not match this app name, the condition does not match.
+	// All the optional fields must match for the rejection condition to apply
```
richvdh, line 103: > 	// All the optional fields must match for the rejection condition to apply

this is the wrong place for this comment: it looks like it's only relevant to `App` (and will be included as part of the generated documentation for `App`). Put it outside the struct please.

main.go
```diff
@@ -97,28 +98,40 @@ type config struct {
...
+	// App name, checked only if not empty
 	App string `yaml:"app"`
-	// Optional: version that must also match in addition to the app and label. If empty, does not check version.
+	// Version, checked only if not empty
 	Version string `yaml:"version"`
-	// Optional: label that must also match in addition to the app and version. If empty, does not check label.
+	// List of labels, checked only if not empty
```
richvdh, line 108: Are you sure it's a list?

main.go
richvdh: I don't think this is fixed. `shouldReject` and `matchesRejectionCondition` both still return a `bool` *and* a `*string`, meaning that they can return nonsensical things like `(false, "blah")`.

Recommend making them return *just* `*string`.

main.go
```diff
@@ -97,28 +98,40 @@ type config struct {
...
+	Reason string `yaml:"reason"`
 }
 
-// shouldReject returns true if the app name AND version AND labels all match the rejection condition.
-// If any one of these do not match the condition, it is not rejected.
-func (c RejectionCondition) shouldReject(appName, version string, labels []string) bool {
-	if appName != c.App {
-		return false
+// shouldReject returns true if all the App, Version, Label and UserTextMatch attributes of a RejectionCondition match
+func (c RejectionCondition) shouldReject(appName, version string, labels []string, userText string) (reject bool, reason *string) {
```
richvdh, line 117: I think it woldl be cleaner to take a `*payload` rather than have this ever-growing list of parameters.

richvdh (changes requested):

main_test.go
```diff
@@ -17,54 +17,84 @@ func TestConfigRejectionCondition(t *testing.T) {
...
 			{
 				App: "block-my-app",
 			},
+			{
+				UserTextMatch: "(\\w{4}\\s){11}\\w{4}",
+				Reason:        "it matches a recovery key and recovery keys are private",
+			},
 		},
 	}
 	rejectPayloads := []payload{
```
richvdh, line 31: could you add a comment here explaining what `Data.ExpectedRejectReason` is about? It seems a reasonable hack but a clarifying comment might be helpful.

main.go
davidegirardi: Done, I also moved the version extraction logic into `shoudReject` from `matchesRejectionCondition` to do it cleanly.

main.go
```diff
@@ -97,55 +98,74 @@ type config struct {
...
+	Reason string `yaml:"reason"`
 }
 
-// shouldReject returns true if the app name AND version AND labels all match the rejection condition.
-// If any one of these do not match the condition, it is not rejected.
-func (c RejectionCondition) shouldReject(appName, version string, labels []string) bool {
-	if appName != c.App {
-		return false
+// shouldReject returns a rejection reason if the payload matches the condition, nil if the server should accept the rageshake
+// gocyclo: ignore we need lots of checks because all fields are optional
```
davidegirardi, line 117: Gocyclo wants a comment `//gocyclo:ignore` on a line by itself to ignore a function. Unfortunately `go fmt` will add a space there `// gocyclo:ignore`. Can I raise the cyclomatic dependency in the linting script from 12 to 13?

richvdh: sorry, forgot to respond to this. (Feel free to "request review", it helps me remember there is something to come back to.)

I'm really not enthusiastic about increasing the cyclomatic complexity limit at all, either locally or globally. Whilst it can feel arbitrary, and to some extent the exact limit *is* arbitrary, it's there for a reason, and is intended to help remind us when it's time to break down a function into smaller parts rather than waiting for a function to get huge. If we increase the limit every time we hit it, we defeat the purpose.

In this case: sorry, but you've broken the camel's back. It's time to split up `shouldReject`. How about a set of functions like

```
func (c RejectionCondition) matchesApp(p *payload) bool {
    // Empty `RejectionCondition.App` is a wildcard which matches anything
    return c.App == "" || c.App == p.AppName
}

func (c RejectionCondition) matchesVersion(p *payload) bool {
	version := ""
	if p.Data != nil {
		version = p.Data["Version"]
	}
    // Empty `RejectionCondition.Version` is a wildcard which matches anything
	return c.Version == "" || c.Version == version
}
```

... and then put those together in the obvious manner in `shouldReject` ?

davidegirardi: I'll think about it but IMO it will make the code harder, not easier, to follow. Let's see where I land.

main.go
```diff
@@ -113,48 +113,71 @@ type RejectionCondition struct {
...
-		if !userTextRegexp.MatchString(p.UserText) {
-			reason = nil
-			return
-		}
-	}
+	reason = c.rejectApp(p, reason)
+	reason = c.rejectLabel(p, reason)
+	reason = c.rejectUserText(p, reason)
+	reason = c.rejectVersion(p, reason)
 	return
```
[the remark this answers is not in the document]
davidegirardi: That was my first approach but every attribute (App, Version, ...) can:
- Match if there's a rejection condition for it and it matches the values (true in your proposal)
- Not match if there's a rejection condition for it but it does not match it (false in your proposal)
- Be irrelevant if there's no condition for it (we'd need to skip it)

I don't know how to express that in a good way with Go and the data structures it provides. For example: if we had lists we could easily express it and build a dynamic `defaultReason` telling you what made the server reject a rageshake at the same time.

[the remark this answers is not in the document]
richvdh: > That was my first approach but every attribute (App, Version, ...) can:
> 
>     * Match if there's a rejection condition for it and it matches the values (true in your proposal)
> 
>     * Not match if there's a rejection condition for it but it does not match it (false in your proposal)
> 
>     * Be irrelevant if there's no condition for it (we'd need to skip it)

Erm. "Match" is the same as "Be irrelevant", isn't it?


```go
func (c RejectionCondition) matchesApp(p *payload) bool {
	// Empty `RejectionCondition.App` is a wildcard which matches anything
	return c.App == "" || c.App == p.AppName
}

func (c RejectionCondition) matchesVersion(p *payload) bool {
	version := ""
	if p.Data != nil {
		version = p.Data["Version"]
	}
	// Empty `RejectionCondition.Version` is a wildcard which matches anything
	return c.Version == "" || c.Version == version
}

func (c RejectionCondition) rejectLabel(p *payload) bool {
	// Empty `RejectionCondition.Label` is a wildcard which matches anything
	if c.Label == true {
		return true
	}

	// Otherwise return true only if there is a label that matches
	labelMatch := false
	for _, l := range p.Labels {
		if l == c.Label {
			labelMatch = true
			break
		}
	}
	return labelMatch
}

func (c RejectionCondition) matchesUserText(p *payload) bool {
    return c.UserTextMatch == "" || regexp.MustCompile(c.UserTextMatch).MatchString(p.UserText)
}

func (c RejectionCondition) shouldReject(p *payload) *string {
	if c.matchesApp(p) && c.matchesVersion(p) && c.matchesLabel(p) && c.matchesUserText(p) {
		// RejectionCondition matches all of the conditions: we should reject this submission/
		defaultReason := "app or user text rejected"
		if c.Reason != "" {
			return &c.Reason
		} else {
			return &defaultReason
		}
	} else {
		return nil
	}
}
```

Untested, but should work I think?

PR #89, opened 2025-02-05
override linear token based on verified user id

KishanBagaria:
this will allow internal users to set their own linear token so issues show up as created by them instead of issue-bot.

(code written entirely by cursor composer)


richvdh:
> edit: pushed to wrong repo

we've all been there :laughing:

PR #90, opened 2025-02-27
Report error responses as JSON

Half-Shot:
For https://github.com/element-hq/element-meta/issues/2732

This changes the following:
 - Errors reported by the application now respond with a JSON body containing both an `error` describing the problem, and an `errcode` which is machine readable.
 - Admins may now (optionally) specify a errcode for each rejection condition.

There is a [docs/api.md](https://github.com/matrix-org/rageshake/blob/c8fdf9639edbc995f9f387675352aeaee5904f45/docs/api.md) to describe the change.

I have done this in a way that works, but I have no doubt this is horrendous Go code. Please be brutal.


docs/api.md
richvdh, line 1: Thank you for taking the time to document this! Note, however, that there exists already API documentation in the readme. I think it makes sense to start a a separate files since it's getting big, but could you move the existing documentation over, and link to the new file from the readme?

main.go
```diff
@@ -54,6 +54,9 @@ var DefaultEmailBodyTemplate string
 var configPath = flag.String("config", "rageshake.yaml", "The path to the config file. For more information, see the config file in this repository.")
 var bindAddr = flag.String("listen", ":9110", "The port to listen on.")
 
+const DefaultErrorReason string = "app or user text rejected"
+const DefaultErrorCode string = "RS_UNKNOWN"
```
richvdh: these things can be private. Golang is awful, so the way you define a private symbol is by starting with a lowercase letter:

```suggestion
const defaultErrorReason string = "app or user text rejected"
const defaultErrorCode string = "RS_UNKNOWN"
```

docs/api.md
```diff
@@ -0,0 +1,31 @@
...
+
+Where the fields are as follows:
+
+ - `error` is an error string to explain the error, in English.
+ - `errcode` is a machine readable error code which can be used by clients to give a localized error.
+ - `policy_url` is an optional URL that links to a reference document, which may be presented to users.
+
+### Error codes
+
+- `RS_UNKNOWN` is a catch-all error when the appliation does not have a specific error.
```
richvdh, line 103: is there an intention that "RS_" be a namespace for server errors, rather than the user-defined ones in `rejection_conditions`? It might be good to mention that explicitly.

main.go
```diff
@@ -148,26 +153,30 @@ func (c RejectionCondition) matchesUserText(p *payload) bool {
 	return c.UserTextMatch == "" || regexp.MustCompile(c.UserTextMatch).MatchString(p.UserText)
 }
 
-func (c RejectionCondition) shouldReject(p *payload) *string {
+func (c RejectionCondition) shouldReject(p *payload) (*string, *string) {
```
richvdh, line 159: could you add a comment saying what the return value means?

main.go
```diff
@@ -148,26 +153,30 @@ func (c RejectionCondition) matchesUserText(p *payload) bool {
...
 	if c.matchesApp(p) && c.matchesVersion(p) && c.matchesLabel(p) && c.matchesUserText(p) {
 		// RejectionCondition matches all of the conditions: we should reject this submission/
-		defaultReason := "app or user text rejected"
+		var reason = DefaultErrorReason
+		if c.Reason != "" {
+			reason = c.Reason
+		}
+		var code = DefaultErrorCode
 		if c.Reason != "" {
```
richvdh: shouldn't this be

```suggestion
		if c.ErrorCode != "" {
```

docs/api.md
richvdh: oh, given your rageshake.sample.yaml, I guess not.

Maybe there *should* be a separate namespace?

Or if not, the prefix seems a bit redundant?

main.go
richvdh: likewise for `matchesRejectionCondition`

submit.go
```diff
@@ -165,14 +165,26 @@ type submitResponse struct {
...
+type submitErrorResponse struct {
+	Error string `json:"error"`
+	ErrorCode string `json:"errcode"`
+}
+
+type submitPolicyErrorResponse struct {
+	Error string `json:"error"`
+	ErrorCode string `json:"errcode"`
+	PolicyUrl string `json:"policy_url,omitempty"`
+}
```
richvdh: do we need both of these? can't we just share one? Which would also mean we can use `writeError` for the policy errors too?

submit.go
```diff
@@ -165,14 +165,26 @@ type submitResponse struct {
...
+
 func (s *submitServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	// if we attempt to return a response without reading the request body,
 	// apache gets upset and returns a 500. Let's try this.
 	defer req.Body.Close()
 	defer io.Copy(ioutil.Discard, req.Body)
 
 	if req.Method != "POST" && req.Method != "OPTIONS" {
-		respond(405, w)
+		writeError(w, 405, submitErrorResponse{"Method not allowed. Use POST.", "RS_METHOD_NOT_ALLOWED"})
```
richvdh, line 187: could you define some `const`s for the error codes, to collect them all together?

richvdh (changes requested):

docs/api.md
Half-Shot: I think there should be a separate namespace, let's do that. Something like `RS_REJECTED_YOUR_REASON_GOES_HERE`

Half-Shot: Documented and tested for during config load :+1:

submit.go
Half-Shot: Have done so. Noting that this is go 1.16 which I believe is before generics are a thing, so I have used what I think is "any struct" which does the trick.

submit.go
Half-Shot: done

errors.go
```diff
@@ -0,0 +1,16 @@
...
+// ErrCodeContentTooLarge is reported when the reported content size is too large.
+const ErrCodeContentTooLarge = "RS_CONTENT_TOO_LARGE";
+// ErrCodeDisallowedApp is reported when a report was rejected due to the report being sent from an unsupported
+const ErrCodeDisallowedApp = "RS_DISALLOWED_APP";
+// ErrCodeMethodNotAllowed is reported when you have used the wrong method for an endpoint.
+const ErrCodeMethodNotAllowed = "RS_METHOD_NOT_ALLOWED";
+// ErrCodeRejected is reported when the submission could be understood but was rejected by RejectionConditions.
+const ErrCodeRejected = "RS_REJECTED";
+// ErrCodeUnknown is a catch-all error when the appliation does not have a specific error.
+const ErrCodeUnknown = "RS_UNKNOWN";
```
richvdh, line 16: missing LF at EOF. Suprised the linter doesn't complain, tbh

richvdh, line 16: you can do:

```
const (
    // ErrCodeBadContent is reported when the reports content could not be parsed.
    ErrCodeBadContent = "RS_BAD_CONTENT"

    // ErrCodeBadHeader is reported when a header was not able to be parsed.
    ErrCodeBadHeader = "RS_BAD_HEADER"

    // etc
}
```

... which is a bit nicer

docs/api.md
```diff
@@ -101,15 +101,15 @@ Where the fields are as follows:
...
 - `RS_BAD_CONTENT` is reported when the reports content could not be parsed.
-- `RS_REJECTED` is is reported when the submission could be understood but was rejected by `rejection_conditions`.
+- `RS_REJECTED` is reported when the submission could be understood but was rejected by `rejection_conditions`.
   This is the default value, see below for more information.
 
 In addition to these error codes, the configuration allows application developers to specify specific error codes
-for report rejection (see the `rejection_conditions` config option). Consult the administrator of your rageshake
-server in order to determine what error codes may be presented.
+for report rejection under the `RS_REJECTED_*` namespace. (see the `rejection_conditions` config option). Consult the
+administrator of your rageshake server in order to determine what error codes may be presented.
```
richvdh, line 115: ok, so what does `RS_` mean here, and do we really need it?

main.go
```diff
@@ -156,14 +154,15 @@ func (c RejectionCondition) matchesUserText(p *payload) bool {
 	return c.UserTextMatch == "" || regexp.MustCompile(c.UserTextMatch).MatchString(p.UserText)
 }
 
+// Returns a rejection reason and error code if the payload should be rejected by this condition, otherwise nil.
```
richvdh, line 157: For a bit more clarity:

```suggestion
// Returns a rejection reason and error code if the payload should be rejected by this condition; otherwise returns `nil` for both results.
```

likewise below

main.go
```diff
@@ -353,5 +353,11 @@ func loadConfig(configPath string) (*config, error) {
 	if err = yaml.Unmarshal(contents, &cfg); err != nil {
 		return nil, err
 	}
+
+	for idx, condition := range cfg.RejectionConditions {
+		if condition.ErrorCode != "" && strings.HasPrefix(condition.ErrorCode, "RS_REJECTED_") == false {
```
richvdh, line 358:
```suggestion
		if condition.ErrorCode != "" && !strings.HasPrefix(condition.ErrorCode, "RS_REJECTED_") {
```

main.go
```diff
@@ -353,5 +353,11 @@ func loadConfig(configPath string) (*config, error) {
...
+			return nil, fmt.Errorf("Rejected condition %d was invalid. ErrorCode must be use the namespace RS_REJECTED_", idx);
```
richvdh, line 359: it's "errorcode" from the admin's point of view, not "ErrorCode"

```suggestion
			return nil, fmt.Errorf("Rejection condition %d was invalid. `errorcode` must be use the namespace `RS_REJECTED_`; got: `%s`", idx, condition.ErrorCode);
```

submit.go
richvdh: We still have definitions of both `submitPolicyErrorResponse` and `submitErrorResponse`. For clarity, I am suggesting that we just do:

```suggestion
type submitErrorResponse struct {
	Error string `json:"error"`
	ErrorCode string `json:"errcode"`
	PolicyURL string `json:"policy_url,omitempty"`
}
```

... and we set PolicyURL to "" where it is unused.

docs/api.md
Half-Shot: RageShake, and it just felt like good form to use a namespace for the error codes. If you have strong feelings, I can remove it.

richvdh: We don't use that abbreviation elsewhere ("rageshake" is not long), and I can't really see the need for a prefix here. Could you remove it?

Half-Shot: Done :+1:

submit.go
```diff
@@ -168,23 +168,19 @@ type submitResponse struct {
...
-}
-
-type submitPolicyErrorResponse struct {
-	Error string `json:"error"`
-	ErrorCode string `json:"errcode"`
 	PolicyURL string `json:"policy_url,omitempty"`
 }
 
 
+
```
richvdh, line 175:
```suggestion
```

submit.go
```diff
@@ -168,23 +168,19 @@ type submitResponse struct {
...
 func (s *submitServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	// if we attempt to return a response without reading the request body,
 	// apache gets upset and returns a 500. Let's try this.
 	defer req.Body.Close()
 	defer io.Copy(ioutil.Discard, req.Body)
 
 	if req.Method != "POST" && req.Method != "OPTIONS" {
-		writeError(w, 405, submitErrorResponse{"Method not allowed. Use POST.", ErrCodeMethodNotAllowed})
+		writeError(w, 405, submitErrorResponse{Error: "Method not allowed. Use POST.", ErrorCode: ErrCodeMethodNotAllowed})
```
richvdh, line 183: You could have just done 

```go
writeError(w, 405, submitErrorResponse{"Method not allowed. Use POST.", ErrCodeMethodNotAllowed, ""})
```

but this is fine

main.go
```diff
@@ -54,6 +54,10 @@ var DefaultEmailBodyTemplate string
 var configPath = flag.String("config", "rageshake.yaml", "The path to the config file. For more information, see the config file in this repository.")
 var bindAddr = flag.String("listen", ":9110", "The port to listen on.")
 
+// defaultErrorReason is the default reason string when not present for a rejection condition
+const defaultErrorReason string = "app or user text rejected"
+
+
```
richvdh, line 60:
```suggestion
```

PR #107, opened 2025-11-06
Add config setting to limit number of log lines in a rageshake

richvdh:


bmarty:
FTR it seems that clients are impacted by this change, see https://github.com/element-hq/element-x-android/issues/5983

stefanceriu:
Seems we missed this memo too, some sort of heads up would've been lovely.

